PR #3, opened 2015-04-06
Mix Dependency Improvements

scrogson:
* Add mix.lock to git
* Set deps environment groups inline
* Make `ibrowse` optional
* Use an empty list as default option for `with` macro

[Review on Reviewable](https://reviewable.io/reviews/monterail/tesla/3)

### DIFF
--- a/lib/tesla/adapter/ibrowse.ex
+++ b/lib/tesla/adapter/ibrowse.ex
@@ -1,63 +1,65 @@
-defmodule Tesla.Adapter.Ibrowse do
-  def start do
-    :ibrowse.start
-  end
-
-  def call(env) do
-    opts = []
-
-    if target = env.opts[:respond_to] do
-
-      gatherer = spawn_link fn -> gather_response(env, target, {:ok, nil, nil, nil}) end
-
-      opts = opts ++ [stream_to: gatherer]
-      {:ibrowse_req_id, id} = send_req(env, opts)
-      {:ok, id}
-    else
-      format_response(env, send_req(env, opts))
+if Code.ensure_loaded?(:ibrowse) do
+  defmodule Tesla.Adapter.Ibrowse do
+    def start do
+      :ibrowse.start
     end
-  end
 
-  defp format_response(env, res) do
-    {:ok, status, headers, body} = res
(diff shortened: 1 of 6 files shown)


lib/tesla/adapter/ibrowse.ex
```diff
@@ -0,0 +1,65 @@
...
+
+          gather_response(env, target, {:ok, _status, _headers, body})
+
+        {:ibrowse_async_response_end, _} ->
+          response = format_response(env, res)
+          send target, {:tesla_response, response}
+      end
+    end
+  end
+end
```
teamon: Does `Code.ensure_loaded?(:ibrowse)` work in compile or run-time?

mix.exs
```diff
@@ -30,18 +30,10 @@ defmodule Tesla.Mixfile do
               "Documentation" => "http://hexdocs.pm/tesla/0.1.0/"}]
   end
 
-  defp deps(:dev) do
-    [{:ibrowse, github: "cmullaparthi/ibrowse", tag: "v4.1.1"},
+  defp deps do
+    [{:ibrowse, github: "cmullaparthi/ibrowse", tag: "v4.1.1", optional: true},
```
teamon: How does `optional: true` work?

teamon: Also I think `exjsx` should not be a hard dependency.

teamon:
Wow, thanks a lot!

I've already merged #2 before seeing this one so I guess a rebase will be needed.

scrogson:
@teamon I've rebased this one on master.

mix.exs
scrogson: `optional: true` sets the dependency is optional and used only to specify requirements. It is left to the user to include the dependency only if they wish to use it.

This is the same way dependencies are setup in `Plug` and `Ecto`. `Ecto` has an optional dependency on `postgrex` and `mariaex` and it is required for the user to include the dependency they wish to use.

I think `exjsx` should eventually not be a dependency at all. A JSON encoder/decoder should be left to the user to include and specify when adding the middleware (Similar to how `Plug` allows the JSON Parser to be configured).

PR #7, opened 2015-05-14
Use request body

ndemonner:
Bodies were being ignored! This fixes that by passing the body down to ibrowse if present.

[Review on Reviewable](https://reviewable.io/reviews/monterail/tesla/7)


teamon:
Ouh, thanks a lot!

PR #12, opened 2016-02-13
Add QueryParams middleware

zph:
First off, I'm excited to see an elixir HTTP library building on the concepts from Faraday. Plug and rack middleware are a powerful concept. Thank you for writing this library.

This PR adds a middleware for including default query params. Does the heavy
lifting of URI manipulation with URI module.

This middleware is helpful because certain APIs require access_tokens in
their query_params and it's helpful to set these as default using a
middlware.

```
defmodule Github.Api do
  use Tesla.Builder

  plug Tesla.Middleware.BaseUrl, "https://api.github.com"
  plug Tesla.Middleware.Headers, %{'User-Agent' => 'elixir'}

  plug Tesla.Middleware.QueryParams, %{access_token: "XXXXXXXXXXXXXXXXXXXXXXX"}
  plug Tesla.Middleware.DecodeJson

  adapter Tesla.Adapter.Ibrowse
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,6 +118,7 @@ end
 
 - `Tesla.Middleware.BaseUrl` - set base url for all request
 - `Tesla.Middleware.Headers` - set request headers
+- `Tesla.Middleware.QueryParams` - set query parameters
 
 ### JSON
 NOTE: requires [exjsx](https://github.com/talentdeficit/exjsx) as dependency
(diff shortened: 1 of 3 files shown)


teamon:
Thanks a lot, and definitely sorry for the long long time with no response.

I'm going to merge this, but it got me thinking that it might be a good idea to revisit the `Tesla.Env` structure and url handling (maybe something better than just string.

zph:
@teamon I'd be quite happy with a non-string representation in ENV :). Thanks for merging this!

PR #13, opened 2016-02-13
Add content type used by Facebook's Graph API for DecodeJSON middleware

zph:
Facebook's graph api uses 'text/javascript' rather than
"application/json".

It makes sense to guard against both of these types.


teamon:
Thanks!

Please see https://github.com/teamon/tesla/commit/4b45a5ad9c3fa0dcb1f3f4b83ad3a447ee9e0e30

zph:
👍

PR #15, opened 2016-03-03
fixed not passthrough opts to ibrowse

zhongwencool:
Now opts only accept   :**respond_to**
but :ibrowse.send_seq/5 support a lot [options](https://github.com/cmullaparthi/ibrowse/blob/master/src/ibrowse.erl#L271)

such as

ModuleClient.get(rul, query, **[connection_timeout: 3000]**)

PS: the default timeout of ibrowse is 30s, it too long............


zhongwencool:
@teamon Thanks for this great project :+1:

PR #17, opened 2016-09-07
Minor cleanups

princemaple:
Hi,

Thanks for building this nice looking library!
While I was reading the source code, I found a few things could be made "better".

Let me know if there is anything you want to keep as is, e.g. explicit call to generate very http verb

Thanks again
Po

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -94,7 +94,6 @@ defmodule Tesla.Builder do
       unquote(generate_api(:put))
       unquote(generate_api(:patch))
 
-      require Tesla.Builder
       import Tesla.Builder, only: [plug: 1, plug: 2, adapter: 1, adapter: 2]
       @before_compile Tesla.Builder
     end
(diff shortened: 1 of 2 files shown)


lib/tesla.ex
```diff
@@ -85,16 +87,8 @@ defmodule Tesla.Builder do
...
-      unquote(generate_api(:delete))
-      unquote(generate_api(:trace))
-      unquote(generate_api(:options))
-      unquote(generate_api(:post))
-      unquote(generate_api(:put))
-      unquote(generate_api(:patch))
+      unquote(generate_http_verbs())
 
-      require Tesla.Builder
       import Tesla.Builder, only: [plug: 1, plug: 2, adapter: 1, adapter: 2]
```
princemaple: when macros are imported, they are automatically required

teamon:
The first version had a lot of macros magic when generating functions so I wanted to keep it as simple as possible, but in this case it is obvious what is being generated.

Thanks!

PR #19, opened 2016-09-07
update deps

### LINKED ISSUE #18: Bump deps, remove git dependency
- ibrowse, it is on hex now
- httparrot, v0.4 has a bug in merging query params, I've sent a PR https://github.com/edgurgel/httparrot/pull/20

princemaple:
close #18 

- [x] ibrowse
- [x] httparrot

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -44,7 +44,7 @@ defmodule Tesla.Mixfile do
   defp deps do
     [
       # http clients
-      {:ibrowse, "~> 4.2.2", optional: true},
+      {:ibrowse, "~> 4.2",   optional: true},
       {:hackney, "~> 1.6.0", optional: true},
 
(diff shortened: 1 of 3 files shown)


teamon:
I think we should go for the lowest version possible - maybe just `>= 4.0.0` would be fine? (Need to check if the API hasn't changed)

princemaple:
Hmm, I don't think it's a good thing to do. Besides, Elixir community is doing fine with SemVer.
You used to have `tag: v4.2`, I'd go with `~> 4.2`. What do you think?

teamon:
Right, I always forget that `~> 4.2` means `4.x`, not `4.2.x`

👍

princemaple:
@teamon this is ready.

PR #21, opened 2016-09-11
Update example dependencies

timbuchwaldt:
The 0.1.0 broke with recent Elixir versions

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Add `tesla` as dependency in `mix.exs`
 
 ```ex
 defp deps do
-  [{:tesla, "~> 0.1.0"},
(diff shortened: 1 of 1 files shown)


teamon:
Thanks a lot!

timbuchwaldt:
Thanks for building Tesla :) Already solved a problem in a super elegant way :) (Passing along Phoenix request-id header to upstream resources as a correlation ID)

teamon:
Glad to hear that :)

Can you share that as an example?

timbuchwaldt:
Sure can: https://gist.github.com/timbuchwaldt/5d089b827bd1c00411465b1169e4e406

PR #22, opened 2016-09-11
Change JSX sample

timbuchwaldt:
Parsing labels to atoms is a bad practise as it can lead to attacks overflowing the atom table and filling the memory with garbage. I switched the example to a different parser option that is safe to use.


teamon:
Right, that was just the first thing that came to my mind. 

I agree we should not promote bad/dangerous practices, thanks!

PR #27, opened 2016-09-20
Some new pull request

teamon:


lib/tesla.ex
```diff
@@ -8,7 +8,7 @@ defmodule Tesla.Env do
   @type url         :: binary
   @type param       :: binary | [{(binary | atom), param}]
   @type query       :: [{(binary | atom), param}]
-  @type headers     :: %{binary => binary}
+  @type headers     :: %{atom => binary}
   @type body        :: any #
```
teamon: This is wrong

lib/tesla.ex
```diff
@@ -8,7 +8,7 @@ defmodule Tesla.Env do
   @type url         :: binary
   @type param       :: binary | [{(binary | atom), param}]
```
teamon: This is bad

PR #28, opened 2016-09-22
Add information about having to add the adapter dependancy.

ootoovak:
As someone new to Phoenix it was something that stumped me for a little while. Hopefully this makes it quicker for the next person. 😄


teamon:
Thanks!

PR #46, opened 2017-01-22
Builder options

amatalai:
Hello,

I added some configuration options for `use Tesla` and `use Tesla.Builder`
- :only and :except - limit generated functions
```
#usage
use Tesla, only: [:get, :post]
use Tesla, only: ~w(delete)a
use Tesla, except: [:get, :post]
use Tesla, except: ~w(delete)a
```
- :docs - disable generating docs, useful when you create package based on Tesla, but don't want docs for it contains docs for all get, post, put etc. functions generated by Tesla
```
defmodule MyProject.Module do
  use Tesla, docs: false

  @doc "custom function"
  def custom_function: get("something")
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,36 @@ GitHub.get("/user/teamon/repos")
 GitHub.user_repos("teamon")
 ```
 
+### Supported options
+
+Tesla.Builder allows to pass following options
+
+#### `:only` and `:except`
+
+Useful when you don't need functions for all http verbs to be generated.
+
+```ex
+  #examples
+  use Tesla, only: ~w(get post)a
+  use Tesla, only: [:delete]
+  use Tesla, except: [:delete, :options]
+```
+
+#### `:docs`
+
+You can disable docs for tesla generated functions if you don't want them to be included in your own project docs.
+
+```ex
+  defmodule MyProject.ApiModule do
+    @moduledoc "Module that does something"
+
+    use Tesla, docs: false
+
+    @doc "Function to get something from somewhere"
+    def custom_function(), do: get(...)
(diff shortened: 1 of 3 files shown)


teamon:
Cool!
I like the `only` & `except` options :+1:
Af for the `doc` flag - I wonder if it wouldn't be better to have a `private` option, so that the `get/post/etc` function would be defined as private - wdyt?

amatalai:
I thought about `:private` as fourth option in next PR, but you are right. `private: true` will be better than `docs: false`, because public functions should have documentation.
I'll try to improve this soon.

amatalai:
Ok, I still see some issues to be resolved

1. 

```ex
defmodule Custom do
  use Tesla, only: [:get], private: true

  def custom_get(), do: get("https://httpbin.org/ip")
end
```
It will produce warnings: `function get/2 is unused`, `function get/3 is unused`, `function request/2 is unused`. We could probably allow Keyword format in `:only` and `:except` options
```ex
use Tesla, only: [get: 1], private: true
```

2.

We cannot disable `request/1` and `request/2` by `:only` or `:except`, because other functions are dependant on them. Maybe we should leave them as public or make them semi-private by `@doc false`.

teamon:
Ahh right :/

Well, the main reason to make these functions private was to skip the docs, so maybe it is just better to have only `doc: false` (so no `private` flag at all as it seems to be bringing more issues than it solves)

amatalai:
done

I have added some examples to readme as well.

teamon:
Awesome!

I'll make a final review today and merge it :)

PR #47, opened 2017-01-22
[WIP] adapter for katipo

amatalai:
At the moment it works fine with osx, and ubuntu trusty. I could not make it work with other linux distros yet.

Let me know if you want to support adapters with such limitations. Maybe I should extract it to separate package tesla_katipo or something?


teamon:
I'm cool with having this adapter in the main repo since we can have it compiled conditionally like hackney or ibrowse :)

What's the deal with `metrics` dependency?

amatalai:
Conflict between hackney and katipo:
screen shot 2017-01-23 at 15 56 20
It cannot be merged in current state, because hex won't allow to publish it like this
```
mix hex.build
** (Mix) Can't build package with overridden dependency metrics, remove `override: true`
```

amatalai:
I have found a workaround (see 23cf602)

Downsides:
- katipo won't be listed as optional dependency in hex
- publishing to hex would require:
```elixir
MIX_ENV=publish mix deps.unlock metrics
MIX_ENV=publish mix deps.get
MIX_ENV=publish mix hex.publish
```
ofc I could create script for this and alias it as `hex.publish` to make it less annoying but first I want to know what do you think about this

teamon:
Hacking hex seems a bit dirty and I'm afraid it could lead to some confusion :|
Let's me ask around and see if there is some other way to do this (apart from creating a separate package for each adapter which I'd really like to avoid).

amatalai:
Ok, it seems to be very rare problem. Maybe it's worth to create issue in hex repo.

teamon:
Hey @amatalai any chance katipo has updated its dependencies?

amatalai:
Hi, unfortunately the problem is that katipo use newer "metrics" than hackney

amatalai:
I don't think it will be ever possible to resolve conflict between erlang packages. Closing it for now.

PR #51, opened 2017-01-28
middleware for data compression

amatalai:
I'm not expert on content-encoding, but it covers my needs in another project, so I decided to make PR


teamon:
Event if it's not perfect (I'm not an expert either) it's a very good starting point :)

PR #54, opened 2017-01-29
Fix warnings in FollowRedirects middleware

KacperPucek:
Elixir was complaining about `env` and `next` being unused.

---
This change is [Reviewable](https://reviewable.io/reviews/teamon/tesla/54)


teamon:
Thanks!

PR #56, opened 2017-02-04
Protobuf middleware

amatalai:


teamon:
I'm not very familiar with protobuf, but it seems like with this middleware you could decode only a single type of message (i.e. only Message1 but not Message2)?

amatalai:
I do not met a situation where single endpoint returns different messages. Usually in this situation there's single message that include other messages inside as `oneof`.

After second thought maybe this is not most useful middleware. I'll close this for now (until someone will have good example why this should be included)

Issue #57, opened 2017-02-06
Set httpc adapter default opts to `autoredirect: false`


teamon:

teamon:
Ref #55

amatalai:
I can take care of this, but it'll be breaking change for anyone using httpc at the moment.

PR #58, opened 2017-02-07
override httpc default autoredirect value

amatalai:


amatalai:
#57 I should probably reference issue from PR not commit in my fork

PR #62, opened 2017-02-11
stream support for ibrowse

amatalai:
Stream support for httpc and ibrowse seems to be similar, maybe I should create Tesla.Shared module or something like that and move some duplicated functions.


gworthe:
@amatalai any progres on this?

teamon:
@amatalai Hey, what's the status of this one?

amatalai:
@teamon hey, I think it's ready for review/merge

PR #71, opened 2017-04-05
Loosen hackney version

### LINKED ISSUE #68: Update dependency on Hackney
Version `1.7` is now released.

blatyo:
Fixes #68


teamon:
Thanks! I changed the dependency to `~> 1.6` so it will accept any `1.x` that is newer or equal `1.6.0`. See 22258eaef8fb2

PR #80, opened 2017-05-24
Add unicode support 

gdub01:
https://github.com/myfreeweb/httpotion/issues/65

Sorry for the messy history.. only 1 line was actually changed. Unicode responses were not getting decoded properly and this seems to fix it.

### DIFF
--- a/lib/tesla/middleware/core.ex
+++ b/lib/tesla/middleware/core.ex
@@ -27,7 +27,7 @@ defmodule Tesla.Middleware.Normalize do
     end
   end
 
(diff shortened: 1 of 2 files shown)


teamon:
Hey, thanks for noticing that. 
Could you add a test case that covers that change? Also, the commit history could be squashed and rebased with current master.

gdub01:
Okay thanks for the heads up on rebasing! That's done and added a test that seems to work, though not sure if it's in the right spot.

Issue #64, opened 2017-02-13
More conventional output


gworthe:
Hello, thanks for this great library.

It seems to me like there's some convention in elixir to make normal function return output as tuple `{:ok, value}` or `{:error, error}` and for function with ! to return only value or raise error.

Would it be difficult to make Tesla work like this?

teamon:
Hi,

Right now all `get/post/put/etc.` functions return either `Tesla.Env` struct or raise an error.

The issue with returning `:ok/:error` tuples is both very hard and trivial.
It's hard because there are many ways the library could behave and many different expectations from its users - there is no simple one-size-fits-all solution. And it's trivial in terms of implementation. 
Please take a look at the two following middlewares:

```ex
defmodule Tesla.Middleware.TupleForGoodResponses do
  @moduledoc """
  Return :ok/:error tuples for "good" responses, i.e. when status code is in the range (200 - 399)
  """
  def call(env, next, _opts) do
    case Tesla.run(env, next) do
      %{status: status} = env when status < 400
        -> {:ok, env}
      env
        -> {:error, env}
    end
  end
end
```

```ex
defmodule Tesla.Middleware.TupleForSuccessfulTransaction do
  @moduledoc """
  Return :ok/:error tuples for successful HTTP transations, i.e. when the request is completed
  (no network errors etc) - but it can still be an application-level error (i.e. 404 or 500)
  """
  def call(env, next, _opts) do
    try do
      {:ok, Tesla.run(env, next)}
    rescue
      %Tesla.Error{} = er ->
        {:error, er}
    end
  end
end
```

As you can see, implementing both scenarios is pretty easy (and we should probably include some kind of these middlewares but definitely with better names).

And since the topic of return values has been brought up here, @michalmuskala got a minute? :)

michalmuskala:
Sorry for the late answer.

An ideal situation would be to have both a raising and a tuple-return version of each function. But, if there are a lot of functions, that's not really sustainable. I really like how the new `ex_aws` is structured - exactly to combat that issue.

Let's look at a `fetch_repos` function for an example GitHub API client. In that architecture, instead of having the  function perform the actual request, it returns some data encoding the request without executing it. This data, can be later passed to a `request/1` or a `request!` function that actually perform the request. This allows having both versions of every function (the raising and non-raising one) with minimal additional code - the error handling is contained and separate from the actual logic.

teamon:
Thank for all the feedback!

[...]

PR #81, opened 2017-05-26
Return error tuple instead of raising errors

gdub01:
I found I have to write extra code to handle the raises. This way the
user gets to choose whether to raise.

When googling around, noticed it was also brought up here:
https://elixirforum.com/t/error-handling-with-tesla/4511/4

### DIFF
--- a/lib/tesla/middleware/core.ex
+++ b/lib/tesla/middleware/core.ex
@@ -6,9 +6,7 @@ defmodule Tesla.Middleware.Normalize do
     |> normalize
   end
 
-  def normalize({:error, reason}) do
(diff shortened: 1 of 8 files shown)


teamon:
This issue has been brought up in the past - https://github.com/teamon/tesla/issues/64
Please mind that changing this behaviour would be a major breaking change.
I'm open to changes, but there must a be good reason for them.

The current behaviour is to return `%Tesla.Env` on successful transaction and raise on adapter/connection errors. As you can is in #64 it is trivial to change this behaviour with a middleware and does not require any changes to the core.

gdub01:
ahh - sorry. Definitely should've thought of the middleware approach. I may close this then as that will do the trick. Ty

PR #82, opened 2017-05-26
allow json decoding to be used with middleware that transforms errors

gdub01:
Using the middleware to handle error raises/tuples along with using json middleware gives this:

** (UndefinedFunctionError) function :error.body/1 is undefined (module :error is not available)
                :error.body({:error, :econnrefused})
        (tesla) lib/tesla/middleware/json.ex:57: Tesla.Middleware.JSON.decodable_body?/1
        

But adding this line handles that case.

### DIFF
--- a/lib/tesla/middleware/json.ex
+++ b/lib/tesla/middleware/json.ex
@@ -53,6 +53,7 @@ defmodule Tesla.Middleware.JSON do
 
   def decodable?(env, opts), do: decodable_body?(env) && decodable_content_type?(env, opts)
 
+  def decodable_body?({:error, _}), do: false 
   def decodable_body?(env) do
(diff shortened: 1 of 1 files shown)


teamon:
If you put your error handling middleware first, i.e. before JSON middleware it should work without this change :)

amatalai:
Btw, I think we could have one most basic error handling middleware here in main repo.
One that would change env to `{:ok, %Env{}}` and errors to `{:error, %Error{}}` should cover the needs of most users and serve as an example for similar custom middlewares.

teamon:
Exactly my thoughts, I should find some time to add that tomorrow :)

gdub01:
Ah yes =) Thanks so much for the great library and fast responses!

PR #87, opened 2017-06-18
Add support for multipart requests.

ijcd:
I needed multipart support so I took a stab at it. Feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,6 +315,8 @@ You can pass a Multipart struct as the body.
 
 Example:
 ```ex
+    alias Tesla.Multipart
+
     mp =
       Multipart.new
       |> Multipart.add_content_type_param("charset=utf-8")
--- a/README.md
+++ b/README.md
@@ -307,3 +307,21 @@ end
 ```
 
 Each piece of stream will be encoded as json and sent as a new line (conforming to json stream format)
+
+
+## Multipart functionality
+
+You can pass a Multipart struct as the body.
+
+Example:
+```ex
+    mp =
+      Multipart.new
+      |> Multipart.add_content_type_param("charset=utf-8")
+      |> Multipart.add_field("field1", "foo")
+      |> Multipart.add_field("field2", "bar", headers: [{:"Content-Id", 1}, {:"Content-Type", "text/plain"}])
+      |> Multipart.add_file("test/tesla/multipart_test_file.sh")
+      |> Multipart.add_file("test/tesla/multipart_test_file.sh", name: "foobar")
+
+    response = Tesla.post("http://httpbin.org/post", mp)
+```
--- a/lib/tesla/multipart.ex
+++ b/lib/tesla/multipart.ex
@@ -1,0 +1,160 @@
+defmodule Tesla.Multipart do
+  @moduledoc """
+  Multipart functionality.
+
+  Example:
+  ```ex
+      mp =
+        Multipart.new
+        |> Multipart.add_content_type_param("charset=utf-8")
+        |> Multipart.add_field("field1", "foo")
+        |> Multipart.add_field("field2", "bar", headers: [{:"Content-Id", 1}, {:"Content-Type", "text/plain"}])
+        |> Multipart.add_file("test/tesla/multipart_test_file.sh")
+        |> Multipart.add_file("test/tesla/multipart_test_file.sh", name: "foobar")
+
+        response = client.post(url, mp)
+   ```
+  """
+
+  @boundary_chars "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789" |> String.split("")
+
+  @type part_stream :: IO.Stream.t | File.Stream.t
+  @type part_value :: String.t | part_stream
+
+  defstruct [
+    parts: [],
+    boundary: nil,
+    content_type_params: []
+  ]
+
+  @type t :: %__MODULE__ {
+    parts: list(Part.t),
+    boundary: String.t,
+    content_type_params: [String.t]
+  }
+
+  defmodule Part do
+    defstruct [
+      body: nil,
+      dispositions: [],
+      headers: [],
+    ]
+
+    @type t :: %__MODULE__ {
+      body: String.t,
+      headers: Keyword.t,
+      dispositions: Keyword.t,
+    }
+  end
+
+  @doc """
+  Create a new Multipart struct to be used for a request body.
+  """
+  @spec new() :: t
+  def new do
+    %__MODULE__{boundary: unique_string(32)}
+  end
+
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -307,3 +307,21 @@ end
...
 
 Each piece of stream will be encoded as json and sent as a new line (conforming to json stream format)
+
+
+## Multipart functionality
+
+You can pass a Multipart struct as the body.
+
+Example:
+```ex
```
amatalai: I think this example is missing `alias Tesla.Multipart`. It could be misleading for someone who read only a readme.

ijcd: Thanks, I'll make the change.

teamon:
Whoo this is really cool! 
I will give it a try after weekend, but the API looks pretty well-thought 👍

README.md
ijcd: Done. Thanks!

ijcd:
I updated the docs with the suggestion. Are there any other changes you would like to see?

ijcd:
Do you require any changes or can we get this merged?

amatalai:
I'm aware that it's been a long time since the PR was created, but it's big one and therefore it requires a lot of free time to review it properly. I'll try over the weekend if other duties won't prevent me.

lib/tesla/multipart.ex
```diff
@@ -0,0 +1,160 @@
...
+  @doc """
+  Add a parameter to the multipart content-type.
+  """
+  @spec add_content_type_param(t, String.t) :: t
+  def add_content_type_param(%__MODULE__{} = mp, param) do
+    %{mp | content_type_params: mp.content_type_params ++ [param]}
+  end
+
+  @doc """
+  Add a field part.
```
amatalai: I would add list of possible opts here

lib/tesla/multipart.ex
```diff
@@ -0,0 +1,160 @@
...
+    part = %Part{
+      body: value,
+      headers: headers,
+      dispositions: [{:name, name}] ++ opts
+    }
+    %{mp | parts: mp.parts ++ [part]}
+  end
+
+  @doc """
+  Add a file part. The file will be streamed.
```
amatalai: same as add_field, also I would add additional example with `:detect_content_type` to moduledoc

amatalai:
I think that list of possible opts should be added to docs for `add_field` and `add_file`.

Also found 2 issues:
- it doesn't work with DebugLogger middleware due to missing pattern match for `%Multipart{}`
- when passed invalid path to `add_file` and then I tried to send request everything hangs, maybe it should check if file exist first

Anyway big 👍 from me

@teamon what do you think about merging this and working on further improvements in separate PRs?

amatalai:
got green light for merge 💚
I'll do it tomorrow. Also, I will copy thing from my previous post to issues to not forget about them

teamon:
There is a to `:mimerl` which is not added to dependencies.

```
warning: function :mimerl.filename/1 is undefined (module :mimerl is not available)
  lib/tesla/multipart.ex:92
```

@ijcd Have you tried https://hex.pm/packages/mime ?

ijcd:
I don't recall if I tried that one. It looks like :mimerl came in with hackney for me, possibly, since benoitc works on both. It's probably better to go with the elixir one as it looks more recently updated and maintained. It's only used for guessing the mime-type on upload.

fdbeirao:
I reached this post because I am trying to use [elixir-google-api](https://github.com/GoogleCloudPlatform/elixir-google-api) and I was messing around with it in the repl/iex.

This is what happened to me: 
> when passed invalid path to add_file and then I tried to send request everything hangs, maybe it should check if file exist first

How is it possible to completely **halt** the repl? Why isn't it "just crashing"?

I now see what I am doing wrong: I am setting "Hello from Elixir!" as the "path".. but it blows my mind that the repl can hang like that, as if it were stuck in some "low-level" `receive()` method.. :( if I was running this code inside an Elixir process, would the process also just hang (_in there_)?

I am running in Windows and these were the commands I run:
```cmd
{:ok, token} = Goth.Token.for_scope("https://www.googleapis.com/auth/cloud-platform")
conn = GoogleApi.Storage.V1.Connection.new(token.token)
GoogleApi.Storage.V1.Api.Objects.storage_objects_insert_simple(conn, "--my--bucket--id--", "multipart", %GoogleApi.Storage.V1.Model.Object{:"contentType" => "plain/text", :"name" => "upload.txt"}, "Hello from Elixir!", [])
```

My intention was to "upload" a file called `upload.txt`, with content `Hello from Elixir!`.. I will now try to find a way to do that. Preferably one that doesn't require writing to temporary files :)

Cheers.

teamon:
Thanks for letting me know elixir-google-api uses tesla! 🎉 

You don't need to create a tmp file - you can use [`add_file_content`](https://hexdocs.pm/tesla/Tesla.Multipart.html#add_file_content/4)

About the hanging - can you check with raw tesla? Does it hang at `add_file` or when perfrming the request?

fdbeirao:
@teamon your suggestion was spot on! I had to locally create a function that uses `add_file_content` behind the scenes, because the original `elixir-google-api` doesn't support that function.

About the freezing thing, it must be something wrong with my setup (I'm looking at you Windows!). 

If I run `iex -S mix` followed by all the Elixir commands that lead to a freeze, and then I `Ctrl+C` to kill it and do it all over again, the second time it works fine 🤦‍♂️ 

I have access to a linux VM, so I'll try too see if it also hangs in there (in there). 

Massive kudos for your hint 👍

teamon:
Happy to help! 🕺 

In case you still have some problems please open a separate issue, good luck!

henry-hz:
@fdbeirao Fabio, pls, how can you uplaod a file into a bucket folder (in case you need to upload to /my-bucket/my-folder/file.txt) ?

fdbeirao:
@henry-hz Henry, in order to place the file in a "folder" I specify the file_name as the full path with folder, separated by `/`. Do keep in mind that [folders are a lie](https://cloud.google.com/storage/docs/naming#object-considerations).

This `file_name` gets added to the query params:
```elixir
defp storage_objects_insert_from_payload(
         connection,
         bucket,
         content_type,
         file_name,
         file_payload
       ),
       do:
         %{}
         |> method(:post)
         |> url("/upload/storage/v1/b/#{bucket}/o")
         |> add_param(:headers, "content-type", content_type)
         |> add_param(:query, :uploadType, "media")
         |> add_param(:query, :name, file_name)
         |> add_param(:body, :body, file_payload)
         |> Enum.into([])
         |> (&Connection.request(connection, &1)).()
         |> decode(%GoogleApi.Storage.V1.Model.Object{})
```

Let me know if this helped you. I haven't touched that code for a while.

PR #93, opened 2017-07-08
Update dependencies and remove deprecation warnings

tompave:
This pull request updates dependencies that either used deprecated Elixir APIs (e.g. `HashDict`) or included other deps in their trees that did.

The Poison dependency was also updated in the `mix.lock` file because it was locked to a very old version.

The PR also updates the `mix.lock` file to the new hex.pm format, which happens automatically when running `mix deps.get` with a recent version of hex. For example:

```diff
-  "cowlib": {:hex, :cowlib, "1.0.2", "9d769a1d062c9c3ac753096f868ca121e2730b9a377de23dec0f7e08b1df84ee", [:make], []},
+  "cowlib": {:hex, :cowlib, "1.0.2", "9d769a1d062c9c3ac753096f868ca121e2730b9a377de23dec0f7e08b1df84ee", [:make], [], "hexpm"},
```

Also dropped support for Elixir 1.3, because the up-to-date dependencies have dropped support for it (Elixir 1.5 is around the corner), and the tests would fail on travis when run with Elixir 1.3.


amatalai:
Hi, 
Thanks for PR. Unfortunately, I don't think it should be merged.

Comparing your branch with master, list of warnings seems to be same for both. (For comparison I created clean mix project using newest elixir 1.5.0-rc.0 and erlang 20.0)
Droping support for 1.3 because of dev/test dependencies doesn't seem to be good idea.
Also mix.lock doesn't make any difference for users, it isn't included in hex package.

tompave:
Hi, thank you for the feedback, but I'm not sure about the comments.

> Comparing your branch with master, list of warnings seems to be same for both.

Can you be more precise, please? If I look at [the latest master build](https://travis-ci.org/teamon/tesla/jobs/250340738) I can see loads of deprecation warnings. Some are for modules that are deprecated in recent versions of Elixir (e.g. `HashDict` and `HashSet`), others are because of style warnings added to the Elixir compiler (e.g. missing parentheses). They all come from the dependencies that I have upgraded. [On the build for this branch](https://travis-ci.org/teamon/tesla/jobs/251600840) there are no warnings.

> For comparison I created clean mix project using newest elixir 1.5.0-rc.0 and erlang 20.0

1.5.0-rc.0 is a pre-release yet. I'm targeting 1.4, the latest stable release.

> Droping support for 1.3 because of dev/test dependencies doesn't seem to be good idea.

That's not the reason I'm proposing to drop it. At the moment this library relies on outdated dependencies that will soon cease to work on modern Elixir releases. I've updated these libraries, which in turn have dropped support for Elixir 1.3. The fact that the Travis tests break with Elixir 1.3 is just a proof that the up-to-date dependencies would not work on modern Elixir, _in production nor in dev_. I mentioned the travis test only to point it out and because it's easy to check the output of the failed build, not to justify dropping 1.3.

> Also mix.lock doesn't make any difference for users, it isn't included in hex package.

I know it's not included in the released package, but it does make a difference.

1. The mix.lock format update is applied by `mix deps.get`, automatically. Anyone who clones the repo today and installs the dependencies will have to deal with it. Since master still contains the old format, the new updated mix.lock file will be seen by git as an update, which is a hassle if people want to work on something more narrow and send a PR. I'm just saying that the new format should be part of master.
2. The mix.lock describes the libraries and versions that are installed when running `mix deps.get`, and as it is now it _locks_ the project to old dependencies. This is an issue because library developers (owners or people sending PRs) will work and run the tests with an old and outdated dependency tree that likely does not reflect the environments used in the wild. For example, at the moment this library is locked to `hackney 1.7.1` and `poison 2.2.0`, which means that the library author develops and tests code against _those versions_, and the same applies to the automated tests in Travis. If people want to use tesla today, however, they will likely use it with the _latest_ versions of those libraries. For example I am using it in a project where I'm requiring `hackney 1.8` and `poison 3.1`. I can use them because Tesla's mixfile allows it (`~> and `>=`), but the fact that the library is developed and tests with old versions gives me zero confidence that it will actually work properly with the current ones.

amatalai:
Ok, I'll try to explain this step by step

Lets create new project
```elixir
mix new foo
```
and add tesla to dependencies in mix.exs
```elixir
defp deps do
    [
      {:tesla, github: "teamon/tesla", branch: "master"}
    ]
end
```
then get deps and compile project. List of compilation warnings for 1.5.0-rc.0 looks like this:
screen shot 2017-07-09 at 10 41 49

Dependencies you are trying to update are dev/test dependencies, they are used only during tesla development to help test things. They aren't included when someone adds tesla to his own project or even listed in hex as part of package. Getting rid of warnings from them is not worth dropping support for one of elixir versions.

tompave:
> Ok, I'll try to explain this step by step

Hey, thanks, but this is not necessary and to be honest it's a bit condescending.
I don't need this explained step by step. I understand what you're saying, don't worry. The problem is still that you're missing the point, and you're trying to explain things that are really not relevant.

> Lets create new project
> ...
> and add tesla to dependencies in mix.exs

I don't know how to reply to this. I guess "... ok, sure".

> then get deps and compile project. List of compilation warnings for 1.5.0-rc.0 looks like this:

Those are all new deprecations [introduced in Elixir 1.5](https://github.com/elixir-lang/elixir/blob/v1.5/CHANGELOG.md#4-deprecations). They have nothing to do with what I'm doing here in this PR. I am trying to remove _old_ deprecation warnings. You're looking at _new, different_ deprecation warnings. And you can see them because you're using a pre-release version of Elixir _that I'm not even considering in this PR_.

> Dependencies you are trying to update are dev/test dependencies, they are used only during tesla development to help test things. They aren't included when someone adds tesla to his own project or even listed in hex as part of package.

Not really. I also updated Poison and hackney in the mix.lock file, which are _runtime dependencies_. I have already explained in my previous message why I think it's important to update them in the lockfile.

With regard to dev/test-only dependencies, I get that as long as the tests run (and test the right things), an outdated test dependency is fine. I still think that one should try to take care of deprecations proactively, rather than waiting for something to break. I appreciate that people can have different opinions on this.

> Getting rid of warnings from them is not worth dropping support for one of elixir versions.

Elixir 1.3 was dropped by Poison, not by a test dependency. I think I've already explained why I think it's a good thing to develop and test the library with up-to-date dependencies.

> I believe there's far more people using elixir 1.3 than people contributing to tesla who can't live with these warnings

Maybe, but that's why you have library versions. People stuck on Elixir 1.3 can keep using Testa 0.7.1, and people on a more modern Elixir can use a newer version of Tesla. I'm not saying to yank (retire) the current versions of Tesla and prevent people on Elixir 1.3 from using it. I'm asking to release a new version.

Also, most people today will start new projects with Elixir 1.4 (soon 1.5), and will want to use recent versions of libraries. Because, you know, new features and bug fixes.

> > At the moment this library relies on outdated dependencies that will soon cease to work on modern Elixir releases.
> They won't cease to work until elixir 2.0

Good to know, thanks.

amatalai:
> Hey, thanks, but this is not necessary and to be honest it's a bit condescending.

Sorry, didn't meant to. I just wanted to be sure that it's clear what I'm comparing.

> Those are all new deprecations introduced in Elixir 1.5. 

Ok, let's leave this. Adding tesla to project with 1.4.x doesn't generate any additional warnings.

> Not really. I also updated Poison and hackney in the mix.lock file, which are runtime dependencies. I have already explained in my previous message why I think it's important to update them in the lockfile.

Diffs on mix.lock are not easiest to read... acknowledged.

> Elixir 1.3 was dropped by Poison, not by a test dependency. I think I've already explained why I think it's a good thing to develop and test the library with up-to-date dependencies.

Still it doesn't seem good reason to drop 1.3 for me. It's optional dependency. Some people may don't need json middleware at all or use exjsx, which allows `elixir: ">= 0.13.3"`. It may be annoying for them.

> Maybe, but that's why you have library versions. People stuck on Elixir 1.3 can keep using Testa 0.7.1, and people on a more modern Elixir can use a newer version of Tesla.
I still see issue here... most PRs are just new middlewares. Let's assume in next month we'll merge support for multiparts. For people using 1.3 this would mean they need to either update elixir for whole project (which probably will cause even more warnings for them) or copy new module from github (which seems poor solution).

tompave:
Hi,

thank you for taking the time to get back to me on this and explain your position.

I understand where you're coming from, but I'm afraid that I'm not really convinced. I could reply with examples of people using this library today or tomorrow, and encountering bugs because they're using it with recent versions of the other libraries, but I guess that you have already considered that.

This is a very useful library and it seems to be working fine for me even with the latest dependencies. So, thanks and keep up the good work.

aphillipo:
It's also worth noting that you can force whichever packages you want across the board with override: true (obviously weird bugs cropping up you are on your own) so I'm really not understanding why this is an issue with the versions made in this package.

PR #100, opened 2017-08-23
safely pass down empty options

leifg:
in case `env.opts` is nil the appending passes down `nil` (`[] ++ nil` evaluates to `nil`)

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -14,7 +14,7 @@ if Code.ensure_loaded?(:hackney) do
         Tesla.build_url(env.url, env.query),
         Enum.into(env.headers, []),
         env.body,
-        opts ++ env.opts
+        (opts ++ List.wrap(env.opts))
       )
     end
     defp request(method, url, headers, %Stream{} = body, opts), do: request_stream(method, url, headers, body, opts)
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -32,7 +32,7 @@ defmodule Tesla.Adapter.Httpc do
       Enum.into(env.headers, [], fn {k,v} -> {to_charlist(k), to_charlist(v)} end),
       content_type,
       env.body,
-      Keyword.split(opts ++ env.opts, @http_opts)
+      Keyword.split(opts ++ List.wrap(env.opts), @http_opts)
     )
   end
 
--- a/lib/tesla/adapter/ibrowse.ex
+++ b/lib/tesla/adapter/ibrowse.ex
@@ -17,7 +17,7 @@ if Code.ensure_loaded?(:ibrowse) do
         Enum.into(env.headers, []),
         env.method,
         body,
-        opts ++ env.opts
+        opts ++ List.wrap(env.opts)
       )
     end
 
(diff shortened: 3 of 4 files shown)


amatalai:
Hi @leifg 

Could you add some tests? Something like:
```elixir
test "basic get request with nil opts" do
    response = B.Client.get("#{http_url()}/ip", opts: nil)
    assert response.status == 200
end
```
inside `Tesla.Adapter.TestCase.Basic` would be perfect.

amatalai:
Also, other adapters may need similar change

leifg:
Good catch, added the test and implemeted the fix for the other adapters

teamon:
@leifg Thanks! 

One question, what is the real use case for this? In tests you explicitly pass `nil`, how can this happen in "real world"? (I mean, you can break almost any code if you explicitly pass nil)

amatalai:
Ah, I was focused on how bad error from this looks (` no function clause matching in Keyword."-split/2-lists^foldl/2-0-"/3`) and didn't though about use case :/

In addition, I checked this second time and same error can be triggered by other non-list truthy values `get(url, opts: 1)`

leifg:
@teamon just saw it pop up in one of my tests suite. It was something along the line of:

`Tesla.get(url, opts: get_opts_from_somewhere_else()`. So basically passing in `nil` implicitly.

leifg:
Is there anything keeping this from being merged?

teamon:
@leifg I'm not convinced this is not a good way to handle nils. You can always simply write 

```ex
Tesla.get(url, opts: get_opts_from_somewhere_else() || [])
```

leifg:
I can see your point.

On the other side it took me a while to find the actual problem, so the error message was hidden and it wasn't clear to me that passing `nil` is the problem. Especially because the return of `++ nil` is not obvious at all.

Also Tesla already handles `nil` explicitly [here](https://github.com/teamon/tesla/blob/master/lib/tesla/adapter/hackney.ex#L6) [here](https://github.com/teamon/tesla/blob/master/lib/tesla/adapter/httpc.ex#L16), [here](https://github.com/teamon/tesla/blob/master/lib/tesla/adapter/ibrowse.ex#L15) and several other locations, so I thought handling `nil` for parsing the env options would make it consistent.

In the end it's up to you if you merge it, no hard feelings.

teamon:
I wonder if we shouldn't add guards instead that would result in better error message (but not perfect).

### Before

```
⌘ ~/code/tesla (master) λ mix test test/tesla_test.exs

  1) test nil options better errors when given nil opts (TeslaTest)
     test/tesla_test.exs:371
     ** (ArgumentError) argument error
     code: Tesla.get("/", nil)
     stacktrace:
       :erlang.++([{:method, :get}, {:url, "/"} | nil], [__module__: Tesla, __client__: nil])
       (tesla) lib/tesla.ex:403: Tesla.perform_request/3
       test/tesla_test.exs:372: (test)
```

### After

```
⌘ ~/code/tesla (master) λ mix test test/tesla_test.exs                                                              1

  1) test nil options better errors when given nil opts (TeslaTest)
     test/tesla_test.exs:371
     ** (FunctionClauseError) no function clause matching in Tesla.get/2

     The following arguments were given to Tesla.get/2:

         # 1
         "/"

         # 2
         nil

     code: Tesla.get("/", nil)
     stacktrace:
       (tesla) lib/tesla.ex:357: Tesla.get/2
       test/tesla_test.exs:372: (test)
```

@leifg, @amatalai wdyt?

teamon:
Required changes https://github.com/teamon/tesla/compare/better-errors

teamon:
Closing this one in favour of #111

PR #97, opened 2017-08-17
logging maps from json

veverkap:
I ran into an error where I was using the JSON middleware AND the debug logging middleware and it couldn't match on `log_body/2` because there was no method that accepted a Map.

I'm not 100% sure this is the best logging.  I'm just using `inspect`

Issue #94, opened 2017-07-15
Make Caching Middlelayer possible


leifg:
I've been playing around with Tesla and I really like it. Great Work!

I am really keen on implementing a caching middleware but I'm kind of lost. By looking at the current code I don't think it's trivial.

The only solution I came up with so far is something like this (very basic).

```
defmodule CachingMiddleware do
  @ttl = 10

  def call(env, next, _options) do
    case Cache.read(env.url) do
      nil -> 
        result = Tesla.run(env, next)
        Cache.write(env.url, result.body, @ttl)
        result
      cached_body ->  %{env | body: cached_body}
    end
  end
end
```

The problem here is that it depending on the cache backend you can't serialize the elixir data structures so you can only take the body, headers and status code.

The following problem is that in the case of a cache hit, the result will be different from the cache miss returns. The only way around this is knowing about the following middlewares (tuples, json, logger) and running them manually when having a cache hit.

I have found no way in the code to set a flag that will essentially avoid calling the adapter.

Any ideas?

teamon:
If you put the cache middleware right before adapter (i.e. last) the other middleware won't matter :)

leifg:
Just tried that and it works! Thank you.

The only thing that slightly bothers me is that all the normalization comes after that.

A great way to fix this is to make it possible to insert the middleware at a specific point, so that the only thing that comes after my caching middleware is the call to the adapter.

teamon:
@leifg please take a look at #104

teamon:
@leifg ping :)

leifg:
Very Nice! That totally helps!

teamon:
You can check wiki for very simple cache example. I started working on proper http cache middleware, but it’s not ready to be shown yet (stay tuned).

venkatd:
Thanks. I got something basic working but I like the approach you have taken. This may be a bit off topic, but would there be a way to persist the raw http response instead of the erlang `:erlang.term_to_binary` format? I am hoping for something more portable and human readable so other systems can take advantage of the same cache.

teamon:
There is a high chance tesla cache will have adapters to adapters but I’m not sure yet. I’d like tesla to deal with http layer and leave storage to something else. Ideas are welcome! (In the separate Cache issue ;) )

teamon:
(This one #189)

leifg:
> @leifg would you mind sharing the code you have for caching? I'm trying to implement this myself :)

[...]

PR #104, opened 2017-09-11
Introduce Tesla.Client with pre/post middleware

teamon:
This change is backward compatible:
- Tesla.build_client API stays the same
- one can still use simple function as a client

Tesla.Client struct can hold any custom `fun` (as before)
but can now hold `pre` and `post` middleware lists.
`pre` works the same way as one given to `build_client`.
`post` middleware is injected right before adapter.
It can be used as a regular middleware but can be also
used to dynamically override adapter in runtime.
This is mostly useful in testing:

```ex
defmodule MyAPI do
  use Tesla

  def help(client \\ %Tesla.Client{}) do
    get(client, "/help")
  end
end

client = Tesla.build_adapter fn env ->
  %{env | status: 200, body: "new body"}
end

assert %{body: "new body"} = MyAPI.get(client)
```

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -476,7 +476,7 @@ defmodule Tesla do
   end
 
   def build_adapter(fun) do
-    %Tesla.Client{fun: fn env, _next -> fun.(env) end}
+    %Tesla.Client{post: [{:fn, fn env, _next -> fun.(env) end}]}
   end
 
   def build_url(url, []), do: url
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -3,7 +3,7 @@ defmodule Tesla.Error do
 end
 
 defmodule Tesla.Env do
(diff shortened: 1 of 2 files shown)


amatalai:
I believe these two should be referenced here #94 #97 :)

teamon:
As far as #94 is concerned there is one more way to do this (and I'm surprised I haven't thought about it earlier)

```elixir
defmodule Cached do
  use Tesla

  adapter :cache

  def cache(env) do
    cond do
      String.ends_with?(env.url, "/cached") ->
        %{env | body: "cached", status: 304}
      true ->
        Tesla.run_default_adapter(env)
    end
  end
end
```

All one needs to do is to write a custom adapter that wraps other adapter. 
I've added `Tesla.run_default_adapter/1,2` as part of this PR to make it more convenient to use the default adapter which can be configured globally.

PR #105, opened 2017-09-11
Allow to pass additional errors to Tuples middleware

### LINKED ISSUE #96: Allow to pass additional errors to `Tesla.Middleware.Tuples`
I'm currently implementing a wrapper for an external API which serves error objects describing any errors that occur on their part. I've written a middleware which checks for this errors and raises a custom error struct containing the error infos from the response.

Since I want to respond with tuples, instead of raising errors I considered using the `Tuples` middleware but it only catches `Tesla.Error`s. It would be great if I could pass my own errors to rescue from to the middlware.

# Example

```elixir
plug Tesla.Middleware.Tuples, rescue_errors: [My.Error]
```

teamon:
Closes #96 

@amatalai @Zeeker wdyt?

### DIFF
--- a/lib/tesla/middleware/tuples.ex
+++ b/lib/tesla/middleware/tuples.ex
@@ -13,10 +13,36 @@ defmodule Tesla.Middleware.Tuples do
         plug Tesla.Middleware.Tuples
         plug Tesla.Middleware.Json
       end
+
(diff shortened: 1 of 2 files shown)


amatalai:
👍

Issue #107, opened 2017-09-17
GET request with HTTP Body


rawkode:
I am trying to make GraphQL requests with Tesla, which requires a body on a GET request. I've tried reading the docs and adapters and calling code I believe should work, but sadly to no avail.

Could you please provide an example of this?

amatalai:
Hi, try this

```elixir
defmodule Client do
  use Tesla
  adapter :hackney

  def get_with_body do
     request(method: :get, url: "http://httpbin.org/anything", body: "foo bar")
  end
end
```

rawkode:
I already tried that with `httpc`, and it failed. I'll add the `hackey` adapter and let you know.

Thanks :+1:

amatalai:
Yeah, `httpc` failed for me too. I think it doesn't support get with body.

rawkode:
Works with `hackey`.

Thanks @amatalai

PR #110, opened 2017-09-24
Introducing Tesla.Mock

teamon:
@amatalai, @hodak please take a look if you have a minute and let me know what do you think.

While this is working, I'm not sure how to handle adapter configuration:
- The default behaviour is to check client, then module (`adapter :hackney` in client module), then `config :tesla, :adapter` and then fallback to `Tesla.default_adapter`
- In test, it would be useful to override all setting and force mock adapter.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ erl_crash.dump
 /doc
 /cover
 .tool-versions
+/sandbox
--- a/README.md
+++ b/README.md
@@ -144,23 +144,40 @@ To use it simply include `adapter :ibrowse` line in your API client definition.
 NOTE: Remember to include ibrowse in applications list.
 
 
-### Test / Mock
+### Testing
 
-When testing it might be useful to use simple function as adapter:
+You can set the adapter to `:mock` in tests.
 
 ```elixir
-defmodule MyApi do
-  use Tesla
+# config/test.exs
+config :tesla, adapter: :mock
+```
 
-  adapter fn (env) ->
-    case env.url do
-      "/"       -> %{env | status: 200, body: "home"}
-      "/about"  -> %{env | status: 200, body: "about us"}
+Then, mock requests before using your client:
+
+
+```elixir
+defmodule MyAppTest do
+  use ExUnit.Case
+
+  setup do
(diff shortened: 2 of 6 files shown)


teamon:
I'm actually thinking about removing the `adapter :hackney` option from module definition and move it into maybe something like:

```ex
config :tesla, MyApp.MyClient, adapter: :hackney
```

This way:
- we can easily migrate old behaviour to new one with single line of change
- keep the flexibility, especially using specific adapters in code using Tesla inside bigger source project
- allow overriding it in tests 
- by removing option to hardcode adapter we can encourage library authors to NOT force specific adapter but leave to choice to the user of a library (That was one of the primary goal at the very beginning of this project, but I think it was missed with the introduction of `adapter` macro)

teamon:
Also, I think about supporting 'Tesla.Mock.mock MyClient, fn ... '

amatalai:
As far as I like idea of Tesla.Mock, I'm not sure about this

> by removing option to hardcode adapter we can encourage library authors to NOT force specific adapter but leave to choice to the user of a library

In one of my project: 
* we had serious issues with `hackney` freezing when user provided some weird url (which propagated to other dependencies using hackney and required whole app restart)
* `httpc` works perfectly fine, but required `[body_format: :binary]` option
* I never used `ibrowse`, I don't know much about it, thus I cannot guarantee it will work properly
* also I cannot guarantee some custom adapters will work properly with my package

also #107 is example when one adapter was working fine where other did not

Maybe it would be better to override `adapter` macro with config instead of replacing it.
It would allow user to change adapter suggested by package author if he really want to, but not crash everything for people who didn't read about required config.

teamon:
The question is, do you force an adapter for a client that is used in different projects? 
In case of private client you could achieve the same with config.

amatalai:
public, it's on hex

teamon:
So, how about we check for the following:

1. check `config :tesla, MyClient, adapter: :foo` setting
2. check for `adapter :foo` in modules code
3. check for `config :tesla, adapter: :foo`
4. use the `Tesla.default_adapter` - this would be `:mock` in test and `:httpc` in other env

@amatalai wdyt?

PR #112, opened 2017-09-30
Docs

teamon:
README & Documentation update.

While writing docs for middlewares I've noticed a bunch of exported functions that should have been private. I've changed most of them to `defp`, but then I started to wonder if that's not a breaking change (well, it is). While tesla built-in middlewares does not use these functions, some other code might rely on that. That's mostly the reason this PR is not yet merged ;)


amatalai:
I wouldn't consider undocumented functions as part of package interface anyway.
I believe I saw something similar in elixir itself, undocumented function was changed to defp.

PR #114, opened 2017-10-03
handle multiparts when file path is invalid

amatalai:
This should fix situation when invalid filepath is given to `Multipart.add_file` and then request is made.

In such cases different adapters behave differently. `:httpc` hangs and other two raise errors.

My proposal is to:
* add `Multipart.add_file!` (bang version) which raises Tesla.Error immediately
* let `Multipart.add_file` mark multipart struct as invalid
* let user handle invalid multipart on its own
* raise Tesla.Error on request with invalid multipart

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -46,6 +46,9 @@ if Code.ensure_loaded?(:hackney) do
     end
     defp request(method, url, headers, %Stream{} = body, opts), do: request_stream(method, url, headers, body, opts)
     defp request(method, url, headers, body, opts) when is_function(body), do: request_stream(method, url, headers, body, opts)
+    defp request(_method, _url, _headers, %Multipart{valid: false}, _opts) do
+      raise Tesla.Error, "could not stream non-existent files"
+    end
     defp request(method, url, headers, %Multipart{} = mp, opts) do
       headers = headers ++ Multipart.headers(mp)
       body = Multipart.body(mp)
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -43,6 +43,9 @@ defmodule Tesla.Adapter.Httpc do
     :httpc.request(method, {url, headers}, http_opts, opts)
   end
 
+  defp request(_method, _url, _headers, _content_type, %Multipart{valid: false}, _opts) do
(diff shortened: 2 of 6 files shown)


teamon:
I'm not yet sure at what point there should be an error raised/returned. Right now I can't think of a use case where one would want to put a non-existing file path into Multipart struct and have it read later when performing a request. I think failing fast here is the best option. And we should probably try to read the file with `File.stream` and act on the return value/expection on that because the file can disappear between check for existence and the actual read.

amatalai:
My initial idea for putting non-existing file into Multipart was to allow this:
```elixir
defmodule Foo do
  use Tesla
  alias Tesla.Multipart

  plug Tesla.Middleware.Tuples

  def bar(path) do
    mp =
      Multipart.new()
      |> Multipart.add_file(path)

    post("...", mp)
  end
end
```
instead of this
```elixir
defmodule Foo do
  use Tesla
  alias Tesla.Multipart

  plug Tesla.Middleware.Tuples

  def bar(path) do
    mp =
      Multipart.new()
      |> Multipart.add_file(path)

    post("...", mp)
  rescue
    ex in Tesla.Error ->
      {:error, ex}
    error ->
      reraise error, System.stacktrace
  end
end
```
as it may be confusing for user, why tuples middleware won't rescue errors from `Multipart.add_file`
However, I will not insist on this 😄 
Maybe it would be better to add note to tuples middleware docs and/or create `Multipart.Error` to distinguish from regular `Error` (from request)

teamon:
With recent changes it will raise `** (File.Error) could not stream "i-do-not-exists.txt": no such file or directory` when reading the stream. 
This is not a HTTP error but a client/user error.

You can always add
```elixir
plug Tesla.Middleware.Tuples, rescue_errors: [File.Error]
```

and have that exception returned as
```elixir
{:error,  %File.Error{action: "stream", path: "i-do-not-exists.txt", reason: :enoent}}
```

amatalai:
It's true for hackney, but ibrowse and httpc behaves differently

c8c18bc7797776d8f88870ccdf2ae5e0 _screen shot 2017-10-07 at 19 20 06

What I wanted to achieve was a consistent behavior for all three of them.

amatalai:
But now I see it's more "how streams are handled by different adapters" kind of problem. Multiparts only made it more noticeable.

You can close this pr, it wouldn't solve whole issue.

PR #117, opened 2017-10-05
Add Multipart.add_file_content function

visciang:
Extend Multipart module with `add_file_content` function.

This allow us to pass a file content data, compared to `Multipart.add_file` where you must have a file on the FS.

Tipical use case: the file content data is retrieved from an external source (ex: DB, cache, ..) or is just in memory.


teamon:
Thanks! I'll review this in the next few days.

teamon:
Merged now, thanks!

PR #120, opened 2017-10-11
Add warning that the shown README is not for the released version but for master

raarts:
This bit me when I was trying to add tests, was scratching my head on why Tesla.Mock was not found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,10 @@
 Tesla is an HTTP client loosely based on [Faraday](https://github.com/lostisland/faraday).
 It embraces the concept of middleware when processing the request/response cycle.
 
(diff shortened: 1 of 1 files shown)


teamon:
The readme is for most recent version - 0.9.0 which does include Tesla.Mock. In fact, there is only a single commit on master after that release.

Please ensure you are using 0.9.0, if there is some issue with Tesla.Mock then this is definitely a bug.

raarts:
Ah, I see that you released 0.9.0 5 days ago. My problem with Tesla.Mock stems from a week ago. Anyway, I think the warning is still valid, and the README missed a link to the docs at the top.

teamon:
Ah, ok. Sure, it makes sense to add this warning. Thanks!

PR #125, opened 2017-10-16
Support for list of maps in normalize_body/1 when is_list

starbelly:
- Add cond statement for handling list of maps in normalize_body/1
   when is list
 - Add supporting test unit for the above condition


starbelly:
The test unit should maybe go somewhere else. Please advise on any other changes you'd like to see in this PR. 

Cheers

PR #128, opened 2017-10-20
Added missing error pattern for Poison 3.x

amatalai:
https://github.com/devinus/poison/blob/3.0.0/lib/poison/parser.ex#L44

rare edge case, but happend to me today


teamon:
Cool! I didn't know you can use different lockfiles 👍 

One thing - I'd move `lockfiles` to `test/lockfiles`, wdyt?

amatalai:
I did not know either until today :D
Sure, already moved

amatalai:
Some time ago, there was discussion that we should test Tesla with newest dependency versions. Now it could be done with blank lockfile or even `LOCKFILE=non-existing`

PR #119, opened 2017-10-06
Merging Globo.com Timeout middle for Tesla

emerleite:
As proposed here: https://github.com/globocom/tesla_timeout/issues/3

If you have any improvements to ask, please tell us, so we can do it and merge.

PR #131, opened 2017-10-24
Don't kill parent process when timeout occurs

edescourtis:
Killing the caller because of a timeout is not what people want to do.

### DIFF
--- a/lib/tesla/middleware/timeout.ex
+++ b/lib/tesla/middleware/timeout.ex
@@ -34,7 +34,12 @@ defmodule Tesla.Middleware.Timeout do
       |> Task.await(timeout)
       |> repass_error
     catch :exit, {:timeout, _} ->
-      Process.exit(task.pid, :kill)
+      if Version.match?(System.version(), ">= 1.4.0") do
+        Task.shutdown(task, 0)
+      else
+        task.pid |> Process.unlink()
+        task.pid |> Process.exit(:kill)
+      end
       raise @timeout_error
     end
   end


lib/tesla/middleware/timeout.ex
```diff
@@ -23,18 +24,29 @@ defmodule Tesla.Middleware.Timeout do
...
-      Process.exit(task.pid, :kill)
+      if shutdown do
+          if Version.match?(System.version(), ">= 1.4.0") do
+            Task.shutdown(task, 0)
+          else
+            Process.unlink(task.pid)
+            Process.exit(task.pid, :kill)
+          end
+      else
+          Process.exit(task.pid, :kill)
```
amatalai: I think you can remove this and whole outer `if`.
In original PR (#119) `Task.shutdown` was replaced by `Process.exit(:kill)` to maintain compatibility with Elixir 1.3. Inner `if` looks like correct solution.

lib/tesla/middleware/timeout.ex
```diff
@@ -23,18 +24,29 @@ defmodule Tesla.Middleware.Timeout do
...
       |> Task.await(timeout)
       |> repass_error
     catch :exit, {:timeout, _} ->
-      Process.exit(task.pid, :kill)
+      if shutdown do
+          if Version.match?(System.version(), ">= 1.4.0") do
+            Task.shutdown(task, 0)
+          else
+            Process.unlink(task.pid)
+            Process.exit(task.pid, :kill)
```
amatalai: It could be done as`task.pid |> Process.unlink() |> Process.exit(:kill)` :)

lib/tesla/middleware/timeout.ex
edescourtis: ok

lib/tesla/middleware/timeout.ex
edescourtis: sure

edescourtis:
Okay that last one should do it

amatalai:
Ah. My fault, sorry. I mistook unlink with other function :(

edescourtis:
Looks good let me know if you need me to change anything.

emerleite:
The PR makes sense. Our problem was because of version < 1.5 behaves wrong using Task.shutdown and we thought Process.exit could do the job, but we forgot to unlink.

PR #133, opened 2017-11-01
Custom client implementation.

### LINKED ISSUE #123: Customizable Client struct module
Hello again! You may remember me from such issues as #122. 😄 

The project I'm working on wraps several APIs. This means that I'm `__using__` Tesla in a few places to construct several API client modules with different behaviours.

Each API client is associated with several modules that actually perform the API calls––the surface of these APIs is very wide and benefits from being split up into conceptual buckets.

It would be nice to have my API-client-calling modules type-check that you are using the correct client for the given module. The cleanest and most idiomatic way to do this would be to pattern match on the client struct type, but all Tesla-built clients are `Tesla.Client` structs. It would be helpful if I could customize the struct for any given call to `use Tesla`.

I envision providing an extra option to the using call, i.e: `use Tesla, client: My.API.Client`. If this parameter is provided it would define a new struct and thread it throughout all pattern matching on its type in Tesla macros. If it is not provided it would thread the existing `Tesla.Client` module reference through instead, keeping everything backwards compatible.

This seems possible based on the current implementation of Tesla––all pattern matching on the module name of the client struct lives within macro calls that could be meta-programmed or parameterized. The few other places it is explicitly referenced we could adjust for, I have some ideas on how to manage. With your blessing I would be happy to pursue this further and submit a PR.

The places we are currently matching on it are within the builder:

- [`Tesla.Builder`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L58)

  - [`defmacro __using__`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L61)

    - [`def request(%Tesla.Client{} = client, options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L107)

  - [`defp generate_api(method, docs) when method in [:post, :put, :patch]`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L205)

    - [`def unquote(method)(%Tesla.Client{} = client, url, body, options) when is_list(options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L219)

    - [`def unquote(method)(fun, url, body, options) when is_function(fun) and is_list(options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L224)

      - [`unquote(method)(%Tesla.Client{fun: fun}, url, body, options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L225)

    - [`def unquote(method)(%Tesla.Client{} = client, url, body)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L241)

    - [`def unquote(method)(fun, url, body) when is_function(fun)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L246)

      - [`unquote(method)(%Tesla.Client{fun: fun}, url, body)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L247)

  - [`defp generate_api(method, docs) when method in [:head, :get, :delete, :trace, :options]`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L277)

    - [`def unquote(method)(%Tesla.Client{} = client, url, body, options) when is_list(options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L291)

    - [`def unquote(method)(fun, url, body, options) when is_function(fun) and is_list(options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L296)

      - [`unquote(method)(%Tesla.Client{fun: fun}, url, body, options)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L297)

    - [`def unquote(method)(%Tesla.Client{} = client, url)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L313)

    - [`def unquote(method)(fun, url) when is_function(fun)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L318)

      - [`unquote(method)(%Tesla.Client{fun: fun}, url)`](https://github.com/teamon/tesla/blob/1976b03af5cf982a4d4103bd8622ebe9be369788/lib/tesla.ex#L319)

christhekeele:
Allows for creating Tesla-usage specific client structs, so multiple APIs within a project can own their own type of client.

Closes #123.

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -41,7 +41,8 @@ defmodule Tesla.Env do
 end
 
 defmodule Tesla.Client do
-  @type t :: %__MODULE__{
+  @type t :: %{
+            __struct__: atom,
             fun:  (Tesla.Env.t, Tesla.Env.stack -> Tesla.Env.t),
             pre:  Tesla.Env.stack,
             post: Tesla.Env.stack
@@ -61,6 +62,10 @@ defmodule Tesla.Builder do
   defmacro __using__(opts \\ []) do
     opts = Macro.prewalk(opts, &Macro.expand(&1, __CALLER__))
     docs = Keyword.get(opts, :docs, true)
+
+    module = Keyword.get(opts, :client, Tesla.Client)
+    generate_client_struct(module)
+    opts = Keyword.put(opts, :client, module)
 
     quote do
       Module.register_attribute(__MODULE__, :__middleware__, accumulate: true)
@@ -104,7 +109,7 @@ defmodule Tesla.Builder do
       else
         @doc false
       end
-      def request(%Tesla.Client{} = client, options) do
+      def request(%unquote(module){} = client, options) do
         Tesla.perform_request(__MODULE__, client, options)
       end
 
@@ -193,16 +198,32 @@ defmodule Tesla.Builder do
     quote do: @__adapter__ {unquote(adapter), unquote(opts)}
   end
 
+  defp generate_client_struct(module) do
+    unless Code.ensure_loaded?(module) do
+      defmodule module do
+        @type t :: %__MODULE__{
+                  fun:  (Tesla.Env.t, Tesla.Env.stack -> Tesla.Env.t),
+                  pre:  Tesla.Env.stack,
+                  post: Tesla.Env.stack
+        }
+        defstruct fun: nil,
+                  pre: [],
+                  post: []
+      end
+    end
+  end
+
   defp generate_http_verbs(opts) do
+    client  = Keyword.fetch!(opts, :client)
     only    = Keyword.get(opts, :only,    @http_verbs)
     except  = Keyword.get(opts, :except,  [])
 
     @http_verbs
     |> Enum.filter(&(&1 in only && not &1 in except))
-    |> Enum.map(&generate_api(&1, Keyword.get(opts, :docs, true)))
+    |> Enum.map(&generate_api(&1, client, Keyword.get(opts, :docs, true)))
   end
 
-  defp generate_api(method, docs) when method in [:post, :put, :patch] do
+  defp generate_api(method, client, docs) when method in [:post, :put, :patch] do
     quote do
       if unquote(docs) do
         @doc """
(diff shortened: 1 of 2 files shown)


teamon:
While I understand your use case I'm not sure this is something that should be added to the core library. This looks like reimplementing type-safe generics in a language that wasn't made to support them.

I can think of one other option to achieve what you need, without changes to tesla itself.

```ex
defmodule TeslaCustomClientTest do
  use ExUnit.Case

  require Tesla
  @url "http://localhost:#{Application.get_env(:httparrot, :http_port)}"

  defmodule ApiA do
    defmodule Client do
      defstruct tesla: nil
    end

    defmodule Driver do
      use Tesla

      # add middlewares here
    end

    def new do
      %Client{tesla: Tesla.build_client([])}
    end

    # write HTTP method wrapers
    def get(%Client{} = client \\ new(), url), do: Driver.get(client.tesla, url)

    # or maybe even better write more descriptive interface
    def items(%Client{} = client), do: Driver.get(client.tesla, "/items")
  end

  test "calls work with custom client client" do
    assert %{status: 200} = ApiA.get(ApiA.new(), @url <> "/")
  end

  test "calls do not accept normal clients" do
    assert_raise FunctionClauseError, fn ->
      ApiA.get(%Tesla.Client{}, @url <> "/")
    end
  end

  test "calls defaults to the custom client" do
    assert %{status: 200} = ApiA.get(@url <> "/")
  end
end
```

All you need to do is to write a wrapper module that will provide the interface you need with desired pattern matching, either via exposing HTTP functions (get, post, etc) or by writing more descriptive interface mapping to specific API endpoints.

teamon:
@christhekeele I'm closing this due to inactivity. Feel free to reopen/comment.

PR #136, opened 2017-11-17
Make Tesla.Mock work outside the test process scope

Zorbash:
The current implementation of `Tesla.Mock` mocks only requests initiated from the test process.
This leads to `Tesla.Mock.mock/1` not being suitable for `setup_all` callbacks or API client modules which perform the requests from within spawned processes.

This PR aims to solve the aforementioned shortcomings in a backwards compatible manner.

### DIFF
--- a/lib/tesla/mock.ex
+++ b/lib/tesla/mock.ex
@@ -92,22 +92,26 @@ defmodule Tesla.Mock do
   @spec mock((Tesla.Env.t -> Tesla.Env.t | {integer, map, any})) :: no_return
   def mock(fun) when is_function(fun) do
     Process.put(__MODULE__, fun)
+    Agent.start_link(fn -> fun end, name: :__tesla_mock__)
   end
 
 
   ## ADAPTER IMPLEMENTATION
 
   def call(env, _opts) do
-    case Process.get(__MODULE__) do
-      nil ->
-        raise Tesla.Mock.Error, env: env
-      fun ->
-        case rescue_call(fun, env) do
-          {status, headers, body} ->
-            %{env | status: status, headers: headers, body: body}
-          %Tesla.Env{} = env ->
-            env
-        end
+    case {Process.get(__MODULE__), Process.whereis(:__tesla_mock__)} do
+      {fun, _pid} when is_function(fun) -> mock_call(fun, env)
+      {nil, pid} when is_pid(pid) -> Agent.get(pid, &(mock_call(&1, env)))
+      _ -> raise Tesla.Mock.Error, env: env
+    end
+  end
+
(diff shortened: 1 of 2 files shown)


teamon:
I see the issue, however the proposed solution is not backward compatible.
Consider the following test cases:

```ex
# test/a_test.exs
defmodule ATest do
  use ExUnit.Case, async: true

  setup_all do
    Tesla.Mock.mock fn
      _env -> %Tesla.Env{status: 200, body: "AAA"}
    end

    :ok
  end

  test "mock get request" do
    assert %Tesla.Env{} = env = Tesla.get("/")
    assert env.body == "AAA"
  end
end
```

and

```ex
# test/b_test.exs
defmodule BTest do
  use ExUnit.Case, async: true

  setup_all do
    Tesla.Mock.mock fn
      _env -> %Tesla.Env{status: 200, body: "BBB"}
    end

    :ok
  end

  test "mock get request" do
    assert %Tesla.Env{} = env = Tesla.get("/")
    assert env.body == "BBB"
  end
end
```

Note, that these test cases both use `async: true`. This means that they will run in parallel.
What happens is:

```bash
λ mix test                                                                                                                                                                                                           
Compiling 3 files (.ex)
==> sandbox
.

  1) test mock get request (ATest)
     test/a_test.exs:12
     Assertion with == failed
     code:  assert env.body() == "AAA"
     left:  "BBB"
     right: "AAA"
     stacktrace:
       test/a_test.exs:14: (test)


Finished in 0.05 seconds
2 tests, 1 failure
```

Test case A got the mock from test case B.
If we add a match

```ex
{:ok, pid} = Agent.start_link(fn -> fun end, name: :__tesla_mock__)
```

There will be an error:

```bash
  1) ATest: failure on setup_all callback, test invalidated
     ** (MatchError) no match of right hand side value: {:error, {:already_started, #PID<0.346.0>}}
     stacktrace:
       (tesla) lib/tesla/mock.ex:95: Tesla.Mock.mock/1
       test/a_test.exs:5: ATest.__ex_unit_setup_all_0/1
       test/a_test.exs:1: ATest.__ex_unit__/2
```

since you can't start two Agents (or any other processes) with the same name.

Zorbash:
I can understands your concerns about `setup_all` callbacks failing for `async: true` mocks. Using `Tesla.Mock`  in `setup_all` is not encouraged / documented / tested and wasn't working even without this PR. In this sense I don't thing we're breaking any current behaviour of the library.

Would it be fair if we logged a warning or failed early when a user sets up a mock from a `setup_all` with `async: true`?

teamon:
Continued in #143

Issue #137, opened 2017-11-23
Put query params from path/url into query list


teamon:
Example:

```
iex(1)> Tesla.get("http://google.com?page=3")
%Tesla.Env{
  __client__: nil,
  __module__: Tesla,
  body: "...",
  headers: %{... },
  method: :get,
  opts: [],
  query: [],
  status: 302,
  url: "http://google.com?page=3"
}
```

The `query` fields is empty, it should be `[page: "3"]`

pawelduda:
Hello @teamon, got a question about this one

How would you see this example URL, taken from tests act?
```
#{@url}/get?page=1&sort=desc&status[]=a&status[]=b&status[]=c&user[name]=Jon&user[age]=20"
 
assert query["page"] == "1"
assert query["sort"] == "desc"
assert query["status[]"]   == ["a", "b", "c"]
assert query["user[name]"] == "Jon"
assert query["user[age]"]  == "20"
```
Currently I wrote a logic for this, utilizing `URI.decode_query/1` and the third example fails with:
```
code:  assert query["status[]"] == ["a", "b", "c"]
left:  "c"
right: ["a", "b", "c"]
```

Should I try something else to keep parity with that example?

teamon:
Let's continue in #139.

PR #139, opened 2017-11-28
Capture GET query params

pawelduda:
This implements #137

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -28,6 +28,8 @@ if Code.ensure_loaded?(:hackney) do
     alias Tesla.Multipart
 
     def call(env, opts) do
+      env = Tesla.Adapter.Shared.capture_query_params(env)
+
       with  {:ok, status, headers, body} <- request(env, opts || []) do
         %{env | status:   status,
                 headers:  headers,
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -28,8 +28,6 @@ if Code.ensure_loaded?(:hackney) do
     alias Tesla.Multipart
 
     def call(env, opts) do
-      env = Tesla.Adapter.Shared.capture_query_params(env)
-
       with  {:ok, status, headers, body} <- request(env, opts || []) do
         %{env | status:   status,
                 headers:  headers,
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -15,7 +15,9 @@ defmodule Tesla.Adapter.Httpc do
   @http_opts ~w(timeout connect_timeout ssl essl autoredirect proxy_auth version relaxed url_encode)a
 
   def call(env, opts) do
+    env = Tesla.Adapter.Shared.capture_query_params(env)
     opts = Keyword.merge(@override_defaults, opts || [])
+
     with {:ok, {status, headers, body}} <- request(env, opts) do
       format_response(env, status, headers, body)
     end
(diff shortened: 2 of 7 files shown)


teamon:
This is awesome, 💃 💃 💃 !

I don't know if you've seen it, but there is a Normalize middleware that unified in/out to/from adapters. Have you though about putting query params extraction inside thhis middleware instead of every adapter?

I also think we should add more unit tests for `capture_query_params/1` function. 

Any reason for limiting query params scraping to `GET` requests only? 
Maybe `capture_query_params/1` could take only the `url` instead of whole `env`? 
(This would make testing easier)

Please let me know if you have any questions.

pawelduda:
All these suggestions make sense, I will take care of these changes as soon as I find some time :+1:.
 
Just one question though, what should be the desired behaviour when someone mixes query params with POST data? How should I capture query params to avoid duplicate keys? Maybe there should be a different key in env variable, just for captured query params, to avoid breaking changes?

teamon:
I suppose the confusion comes from web frameworks that tend to mix query params & data extracted from body. In the HTTP spec, "query" is only the part of URL after "?". HTTP itself does not care much about the body of a request - it can be encoded in many different ways, like `JSON`, `XML` or the one used by standard browser forms `form-urlencoded`. 

You can play with curl and httpbin.org to see what is what:

```
~ λ curl -X POST -v "https://httpbin.org/post?a=123&b=456" -d "a=X&b=Y"
# REQUEST HEADERS
> POST /post?a=123&b=456 HTTP/1.1
> Host: httpbin.org
> User-Agent: curl/7.54.0
> Accept: */*
> Content-Length: 7
> Content-Type: application/x-www-form-urlencoded
>
# REQUEST BODY
a=X&b=Y

# RESPONSE HEADERS
< HTTP/1.1 200 OK
< Connection: keep-alive
< Server: meinheld/0.6.1
< Date: Mon, 11 Dec 2017 12:02:38 GMT
< Content-Type: application/json
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
< X-Powered-By: Flask
< X-Processed-Time: 0.00199604034424
< Content-Length: 443
< Via: 1.1 vegur
<
# RESPONSE BODY
{
  "args": {
    "a": "123",
    "b": "456"
  },
  "data": "",
  "files": {},
  "form": {
    "a": "X",
    "b": "Y"
  },
  "headers": {
    "Accept": "*/*",
    "Connection": "close",
    "Content-Length": "7",
    "Content-Type": "application/x-www-form-urlencoded",
    "Host": "httpbin.org",
    "User-Agent": "curl/7.54.0"
  },
  "json": null,
  "origin": "83.19.157.237",
  "url": "https://httpbin.org/post?a=123&b=456"
}
```

httpbin.org returns the query params and body params separately as "args" and "form" JSON keys.

Getting back to your question - since query and body are two different things, even if often treated as one, in case of Tesla's env they are separate so there is no need to worry about them mixing :)

pawelduda:
@teamon Sorry for being absent for so long. I am now back to finish this :)

What I did:
- moved logic for capturing query params into the `Normalize` middleware. While this is an improvement, I am not sure if that logic belongs in this very middleware. Perhaps we could create a separate middleware which only takes care of capturing query params? What do you think?

When it comes to more unit tests, I have basically copied the cases from integration tests but I am currently having trouble with being more creative here. If you have ideas for any specific unit test cases, that you feel are missing, let me know.

I also decided to have the function take the whole env as an argument as I saw, that other modules in `middleware/core.ex` behave in a similar manner. I don't think it impacts testing too much in this case.

teamon:
Welcome back @pawelduda! 

The implementation and tests looks good 🥇 
In the meantime I've started working on 1.0 version (in a `1.0` branch). 
In that branch the Normalize middleware is completely removed, however I'm not yet sure where to fit this PR in the new tesla - I will need some more time to figure this out.

Thanks! (And stay tuned ;))

teamon:
This one is trickier then I thought, there are many cases when things could go wrong. I need some more time to do this properly.

teamon:
I'm still not sure how this one should work, but the current implementation (master) seems to be the easiest to reason about. 

I also don't see much interest in this feature. 
Please feel free to reopen.

Issue #140, opened 2017-11-30
Mock not working properly?


gottfrois:
Hey guys,

Following the readme, I've added the following mock:

```elixir
setup do
  Tesla.Mock.mock fn
    %{method: :post, url: "http://foo.com"} ->
      %Tesla.Env{status: 200, body: "hi!"}
  end

  :ok
end
```

Using the following abstraction:

```elixir
defmodule Notifications.Webhooks.Http do
  use Tesla, only: [:post]

  plug Tesla.Middleware.Tuples
  plug Tesla.Middleware.Retry, delay: 1_000, max_retries: 5
  plug Tesla.Middleware.FollowRedirects
  plug Tesla.Middleware.Timeout, timeout: 5_000
  plug Tesla.Middleware.Logger

  def send(url, headers, body) do
    case post(url, body, headers: headers) do
      {:ok, env}        -> {:ok, %{status: env.status, body: env.body, headers: env.headers}}
      {:error, reason}  -> {:error, reason.message}
    end
  end
end
```

But for some reasons, the test fails with:

```
1) test successfuly sending a webhook should create a new delivery (Notifications.WebhookTest)
     test/notifications/webhooks/webhook_test.exs:25
     ** (Tesla.Mock.Error) There is no mock set for process #PID<0.321.0>.
     Use Tesla.Mock.mock/1 to mock HTTP requests.

     See https://github.com/teamon/tesla#testing

     code: Notifications.send(@webhook)
     stacktrace:
       (tesla) lib/tesla/middleware/timeout.ex:55: Tesla.Middleware.Timeout.repass_error/1
       (tesla) lib/tesla/middleware/timeout.ex:35: Tesla.Middleware.Timeout.call/3
       (tesla) lib/tesla/middleware/follow_redirects.ex:40: Tesla.Middleware.FollowRedirects.redirect/3
       (tesla) lib/tesla/middleware/retry.ex:39: Tesla.Middleware.Retry.retry/4
       (tesla) lib/tesla/middleware/tuples.ex:41: Tesla.Middleware.Tuples.call/3
       (notifications) lib/notifications/webhooks/http.ex:11: Notifications.Webhooks.Http.send/3
       (notifications) lib/notifications/webhooks/webhook.ex:19: Notifications.Protocol.Notifications.Webhooks.Webhook.send/1
       test/notifications/webhooks/webhook_test.exs:26: (test)
```

I'm using `v0.9.0`.

pawelduda:
[...]

PR #143, opened 2017-12-07
Global mock support

teamon:
Continuation of #136, also related to #140

I've moved setting up global mock into separate function to be explicit about it. It will also fail fast when used inappropriately with async test cases.

@Zorbash, @amatalai please take a look.

@gottfrois, @pawelduda, @Daniel-Xu please try using tesla from `global-mock` branch and let us know if there are any issues.

```elixir
{:tesla, github: "teamon/tesla", branch: "global-mock"}
```

### DIFF
--- a/lib/tesla/mock.ex
+++ b/lib/tesla/mock.ex
@@ -57,6 +57,38 @@ defmodule Tesla.Mock do
     %{method: :post} -> {201, %{}, %{id: 42}}
   end
   ```
+
+  ### Global mocks
+  By default, mocks are bound to the current process,
+  i.e. the process running a single test case.
+  This design allows proper isolation between test cases
+  and make testing in parallel (`async: true`) possible.
+
+  While this style is recommended, there is one drawback:
+  if Tesla client is called from different process then the
+  test case it will not get proper mock.
+
+  To solve this issue it is possible to setup a global mock
+  using `mock_global/1` function.
+
+  ```
+  defmodule MyTest do
+    use ExUnit.Case, async: false # must be false!
+
+    setup_all do
+      Tesla.Mock.mock_global fn
+        env -> # ...
+      end
+
+      :ok
+    end
+
+    # ...
+  end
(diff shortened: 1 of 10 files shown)


lib/tesla/mock.ex
```diff
@@ -88,17 +120,27 @@ defmodule Tesla.Mock do
...
   @spec mock((Tesla.Env.t -> Tesla.Env.t | {integer, map, any})) :: no_return
-  def mock(fun) when is_function(fun) do
-    Process.put(__MODULE__, fun)
-  end
+  def mock(fun) when is_function(fun), do: pdict_set(fun)
+
+  @doc """
+  Setup global mocks.
+
+  **WARNING**: This mock will be available to ALL process.
```
Zorbash: typo: processes

lib/tesla/mock.ex
```diff
@@ -57,6 +57,38 @@ defmodule Tesla.Mock do
...
+
+  ### Global mocks
+  By default, mocks are bound to the current process,
+  i.e. the process running a single test case.
+  This design allows proper isolation between test cases
+  and make testing in parallel (`async: true`) possible.
+
+  While this style is recommended, there is one drawback:
+  if Tesla client is called from different process then the
+  test case it will not get proper mock.
```
Zorbash: I suggest a minor rewording to:
> If Tesla client is called from a different process then the test case will not be mocked.

lib/tesla/mock.ex
teamon: 👍

lib/tesla/mock.ex
teamon: The test case process itself will be mocked. Either way, both messages would be wrong ;)

It should be something like this "if Tesla client is called from different process it will not use the setup mock"

Zorbash: Looks good :-)

PR #148, opened 2017-12-28
Use mox as Tesla.Mock implementation

teamon:
At this stage it is just an experiment, let's see if it works :)

You can test it yourself using tesla from `mox` branch in your project dependencies:

```ex
{:tesla, github: "teamon/tesla", branch: "mox"}
```

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -56,6 +56,10 @@ defmodule Tesla.Middleware do
   @callback call(env :: Tesla.Env.t(), next :: Tesla.Env.stack(), options :: any) :: Tesla.Env.t()
 end
 
+defmodule Tesla.Adapter do
+  @callback call(env :: Tesla.Env.t(), options :: any) :: Tesla.Env.t()
+end
+
 defmodule Tesla.Builder do
   @http_verbs ~w(head get delete trace options post put patch)a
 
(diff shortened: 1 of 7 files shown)


teamon:
Tests are failing because mox only supports elixir 1.5+
Also, `lib/tesla/mock.ex` should probably only be compiled when `Mix.env == :test`

amatalai:
> Also, lib/tesla/mock.ex should probably only be compiled when Mix.env == :test

I don't think it could work outside tesla project. Deps added in mix.exs are compiled as `:prod`
Tested with [{:deps_env, github: "amatalai/deps_env", only: :dev}](https://github.com/amatalai/deps_env/blob/master/lib/deps_env.ex)

teamon:
Are you 100% sure? I haven't tested that yet, but I always got the impression the dependencies are compiled for every mix env separately since there can be configuration changes.

amatalai:
Petty sure
```elixir
MIX_ENV=dev iex -S mix
iex(1)> DepsEnv.env()
:prod
```
I also don't have access to module placed in `test/support` despite
```elixir
defp elixirc_paths(:dev), do: ["lib", "test/support"]
defp elixirc_paths(:test), do: ["lib", "test/support"]
defp elixirc_paths(_), do: ["lib"]
```
in dependency mix.exs

I'm not 100% sure how mix handle dependency configs but it seems that it compiles dependency as `:prod` with configs from main project `Mix.env()`

teamon:
Closing this, it's not worth the hassle for now.

PR #149, opened 2018-01-01
Improve Tesla.Mock test with JSON middleware

visciang:
I think this is more specific (and clear) given that `MockClient` plugs in `Tesla.Middleware.JSON`.


visciang:
@stefanchrobot I landed here for the same reason

> I forgot to add the Content-Type header to the response and the body was not deserializing to JSON

teamon:
👍 👍 👍

PR #166, opened 2018-02-11
Removing useless rescue clause on Timeout middleware

HugoLnx:
Just removing a useless `rescue` clause from Timeout middleware. When I coded this method in the project I worked with @emerleite, I didn't know that `catch` also works for raised exceptions.

PS.: No test added because this is a refactoring, this code are already tested on: [timeout_test.exs#L99](https://github.com/teamon/tesla/blob/master/test/tesla/middleware/timeout_test.exs#L99)

### DIFF
--- a/lib/tesla/middleware/timeout.ex
+++ b/lib/tesla/middleware/timeout.ex
@@ -51,9 +51,6 @@ defmodule Tesla.Middleware.Timeout do
     Task.async(fn ->
       try do
         {:ok, func.()}
-      rescue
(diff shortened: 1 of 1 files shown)


emerleite:
@HugoLnx makes sense. LGTM

teamon:
Thanks! I will apply these changes to the 1.0 branch.

teamon:
Fixed during other changes so this is not needed anymore. Thanks!

Issue #109, opened 2017-09-20
Dialyzer can't detect tuple when middleware Tuples is used


ahmadferdous:
Hi,

I have a module that uses Tesla and its Tuples middleware.
```
  use Tesla
  plug Tesla.Middleware.Tuples
```

In that module, Tesla functions are used in the following way:
```
    login_url
    |> post(%{username: username, password: password, format: "json"})
    |> extract_body
```

Spec of `extract_body` expects a tuple with the first element `:ok` or `:error`. However, when I run `mix dialyzer`, it gives the following error:
```
The pattern {'ok', _response@1} can never match the type 
#{
    '__client__':=fun(), 
    '__module__':=atom(), 
    '__struct__':='Elixir.Tesla.Env', 
    'body':=_, 
    'headers':=#{binary()=>binary()}, 
    'method':='delete' | 'get' | 'head' | 'options' | 'patch' | 'post' | 'put' | 'trace', 
    'opts':=[any()], 
    'query':=[{_,_}], 
    'status':=integer(), 
    'url':=binary()
}
```

In order for dialyzer to detect that tuples will be returned, do I need to do something additional in the spec or code? What am I missing here?

teamon:
The return type spec for `get/post/...` is `Tesla.Env.t`. When using Tuples middleware this is changed to `{:ok, Tesla.Env.t} | {:error, any}` but this is not reflected in the typespec, hence the dialyzer error.

I can think of some way to check for Tuples middleware during compilation and adjust the generated typespecs but that would require figuring out a way to make this a general feature (i.e. one could make a custom middleware with custom return values that would need to reflected in the generated typespec).

amatalai:
I have workaroud for now: `use Tesla, docs: false`

I had no knowledge of dialyzer when I wrote `:only/:except/:docs` options... `docs: false` unintentionally disables typespecs as well, so you can write your own inside your module.

ahmadferdous:
`use Tesla, docs: false` worked like a charm!

I would like to keep this issue open for a few days, hoping to get more feedback on this. @teamon , is it okay?

amatalai:
Definitely, it should stay open. I should open this myself ~ two weeks ago 😞

teamon:
Alright, it can be done.

### Example usage
```elixir
defmodule Sandbox do
  defmodule Client do
    use Tesla
  end

  defmodule TuplesClient do
    use Tesla

    plug Tesla.Middleware.Tuples
  end

  def main do
[...]

PR #181, opened 2018-02-21
update type spec for middleware

BenMorganIO:
Offense:

```erlang
lib/tesla/middleware/tuples.ex:40: The inferred return type of call/3 ({'error',_} | {'ok',_}) has nothing in common with #{'__client__':=fun(), '__module__':=atom(), '__struct__':='Elixir.Tesla.Env', 'body':=_, 'headers':=#{binary()=>binary()}, 'method':='delete' | 'get' | 'head' | 'options' | 'patch' | 'post' | 'put' | 'trace', 'opts':=[any()], 'query':=[{atom() | binary(),binary() | [{atom() | binary(),binary() | [{_,_}]}]}], 'status':=integer(), 'url':=binary()}, which is the expected return type for the callback of the 'Elixir.Tesla.Middleware' behaviour
```

The return types from `Tesla.Middleware` was only `Tesla.Env.t`. However, `Tesla.Middleware.Tuples` was returning `{:ok, Env.t} | {:error, any()}`. I've updated the return spec for this. While I was doing this, I updated some of the docs; line endings and s/transations/transactions.

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -52,10 +52,6 @@ defmodule Tesla.Client do
             post: []
 end
 
-defmodule Tesla.Middleware do
-  @callback call(env :: Tesla.Env.t(), next :: Tesla.Env.stack(), options :: any) :: Tesla.Env.t()
-end
-
 defmodule Tesla.Builder do
   @http_verbs ~w(head get delete trace options post put patch)a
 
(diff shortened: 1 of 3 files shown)


teamon:
Hi again :)

I appreciate the changes but please keep them to the minimum - I am working on a `1.0` branch and I'd like to have as few merge conflicts as possible.

The `Tuples` middleware is a special one and this is a known type error (see #109). 
I'd rather keep it as a exception and not change the `Tesla.Middleware` spec.

teamon:
@BenMorganIO I'm closing this. Please take a look at `1.0` branch as it solves this issue.

PR #182, opened 2018-02-23
add option to log request duration

BenMorganIO:
So, over at the company I for, we decided to check Sentry and discovered that when there was an error, Sentry would count each error different based on the message. And the message changed because of the duration. This config option allows messages to, more or less, stay the same for most endpoints.

This solves logging issues for `/api/orders.json`, but not for `/api/orders/R123456789.json`.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -2,7 +2,9 @@
 # and its dependencies with the aid of the Mix.Config module.
 use Mix.Config
 
-config :tesla, adapter: :httpc
+config :tesla,
+  adapter: :httpc,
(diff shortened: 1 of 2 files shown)


teamon:
Hi, thanks for reporting this.
Middlewares are supposed to be as independent from the core as possible, and because of that I can not accept this PR adding a new configuration parameter.

To solve your case it would be best to use a custom logger middleware.

On the other hand, I'm not sure how is reporting exceptions with Sentry related to Logger output, do you report every log line as separate exception?

PR #185, opened 2018-02-28
Customize log levels in Logger middleware

nicolasgarnil:
## Problem

The `Logger` middleware has fixed logger levels for different HTTP status codes. This is an issue when, for example, 404 responses are expected and should be logged with `info` level instead of with `error` level.

## Solution

Be able to pass custom status codes to the middleware.

E.g:

```elixir
defmodule MyClient do
  use Tesla

  plug Tesla.Middleware.Logger, %{404 => :info}
end
```

### DIFF
--- a/lib/tesla/middleware/logger.ex
+++ b/lib/tesla/middleware/logger.ex
@@ -15,6 +15,15 @@ defmodule Tesla.Middleware.Logger do
   end
   ```
 
+  ### Customize logging levels
+  ```
+  defmodule MyClient do
+    use Tesla
+
+    plug Tesla.Middleware.Logger, [{200, :info}, {300, :warn}, {400, :info}, {500, :error}]
+  end
+  ```
+
   ### Logger output
   ```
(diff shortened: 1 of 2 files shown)


teamon:
I see the issue, but this implementation is too specific. What we could do is:

- Add a configuration option, but name it as `log_level` (so there is `plug Tesla.Middleware.Logger, log_level: here`
- Instead of map, allow passing a function `(Tesla.Env.t) -> :info | :warn | :error | :debug |:nolog`
- Provide a default, public `default_log_level` function that implement the current behaviour.

Example full usage:

```ex
  plug Tesla.Middleware.Logger, log_level: &my_log_level/1

  def my_log_level(env) do
    case env.status do
      404 -> :info
      _ -> Tesla.Middleware.Logger.default_log_level(env)
    end
  end
```

@nicolasgarnil what do you think? 
/cc @amatalai

nicolasgarnil:
Makes total sense!

If we are all ok with your proposal I can submit a new PR.

teamon:
@nicolasgarnil Let's do it!

PR #201, opened 2018-05-08
add infinity retry option

Ljzn:
In some cases, infinite retries are useful.


teamon:
@Ljzn Thanks! Having infinity option does makes sense :)

Please take a look at Travis CI output - one build has failed because the code is not formatted.
Also, we need to provide some tests for `:infinity`

teamon:
Hey @Ljzn, how's going? :)

teamon:
@Ljzn ping ping

Ljzn:
#222

PR #203, opened 2018-05-15
[WIP] Update the redirect middleware to preserve the original request

mpinkston:
The current `Tesla.Middleware.FollowRedirects` simply updates the url of the response env and tries again. This means that the body from the previous response along with the headers are sent in the next request which can cause unpredictable results.

This PR preserves the original request for each redirect so the body (and optionally headers) can be used in the next request. 

Additionally
- added support for 303 redirects
- add the option to rewrite the method to `:get` on 301 and 302 responses
- optionally prevent or allow redirecting on 301 and 302 responses if the original request was other than `:get` or `:head`

I haven't written tests for all the cases yet, but if it looks good, please let me know and I'd be happy to write them (and fix any bugs I find while doing so)

### DIFF
--- a/lib/tesla/middleware/follow_redirects.ex
+++ b/lib/tesla/middleware/follow_redirects.ex
@@ -15,45 +15,118 @@ defmodule Tesla.Middleware.FollowRedirects do
 
   ### Options
   - `:max_redirects` - limit number of redirects (default: `5`)
+  - `:force_redirect` - If the server response is 301 or 302, proceed with the redirect even
+                        if the original request was neither GET nor HEAD. Default is `false`
+  - `:rewrite_method` - If the server responds is 301 or 302, rewrite the method to GET when
+                        performing the redirect. This will always set the body to nil.
+                        Default is `false`
+  - `:preserve_headers` - Preserve the headers from the original request and send them along in the
+                          redirect. Default is `false`
 
   """
 
   @max_redirects 5
-  @redirect_statuses [301, 302, 307, 308]
+  @force_redirect false
+  @rewrite_method false
+  @preserve_headers false
 
   def call(env, next, opts \\ []) do
-    max = Keyword.get(opts || [], :max_redirects, @max_redirects)
+    opts =
+      opts
+      |> Keyword.put_new(:max_redirects, @max_redirects)
+      |> Keyword.put_new(:force_redirect, @force_redirect)
+      |> Keyword.put_new(:rewrite_method, @rewrite_method)
+      |> Keyword.put_new(:preserve_headers, @preserve_headers)
 
-    redirect(env, next, max)
+    # Initial value for remaining attempts
+    rem = Keyword.fetch!(opts, :max_redirects)
+
+    process_request(env, next, opts, rem)
   end
 
-  defp redirect(env, next, left) when left == 0 do
-    case Tesla.run(env, next) do
-      {:ok, %{status: status} = env} when not (status in @redirect_statuses) ->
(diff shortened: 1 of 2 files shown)


teamon:
To use the original request all you need is to change one variable name in three places:

```diff
diff --git a/lib/tesla/middleware/follow_redirects.ex b/lib/tesla/middleware/follow_redirects.ex
index e835c86..f7ae117 100644
--- a/lib/tesla/middleware/follow_redirects.ex
+++ b/lib/tesla/middleware/follow_redirects.ex
@@ -42,13 +42,13 @@ defmodule Tesla.Middleware.FollowRedirects do

   defp redirect(env, next, left) do
     case Tesla.run(env, next) do
-      {:ok, %{status: status} = env} when status in @redirect_statuses ->
-        case Tesla.get_header(env, "location") do
+      {:ok, %{status: status} = res} when status in @redirect_statuses ->
+        case Tesla.get_header(res, "location") do
           nil ->
-            {:ok, env}
+            {:ok, res}

           location ->
-            location = parse_location(location, env)
+            location = parse_location(location, res)
             redirect(%{env | url: location}, next, left - 1)
         end

```

I agree this is something that should be fixed.

After all your changes the code is a bit complex; the purpose of built it middlewares is to provide the most common scenarios and a an examples for custom middlewares, so I'm careful with adding many options. Having said that I will be more than happy to discuss these features individually.

mpinkston:
Thanks for the feedback @teamon. 
I think I can get behind the idea of keeping packaged middlewares as simple as possible, and if the idea is to provide examples of what the end user can implement in custom use-cases, then even more so. 

Considering that, I might have gone a bit far in trying to strictly stick to the RFC and allow configuration where there was ambiguity (though my specific use-case goes so far as to require session and cookie management as well, spiraling way out of scope there).
So, at the very least I'd like to preserve that original request for the obvious reasons and if possible, add support for the 303 if that seems like it might be common enough.

Perhaps something like this?
```diff
@@ -19,7 +19,7 @@
   """

   @max_redirects 5
-  @redirect_statuses [301, 302, 307, 308]
+  @redirect_statuses [301, 302, 303, 307, 308]

   def call(env, next, opts \\ []) do
     max = Keyword.get(opts || [], :max_redirects, @max_redirects)
@@ -42,13 +42,17 @@

   defp redirect(env, next, left) do
     case Tesla.run(env, next) do
-      {:ok, %{status: status} = env} when status in @redirect_statuses ->
-        case Tesla.get_header(env, "location") do
+      {:ok, %{status: status} = res} when status in @redirect_statuses ->
+        case Tesla.get_header(res, "location") do
           nil ->
-            {:ok, env}
+            {:ok, res}
+
+          location when status == 303 ->
+            location = parse_location(location, res)
+            redirect(%{env | method: :get, url: location}, next, left - 1)

           location ->
-            location = parse_location(location, env)
+            location = parse_location(location, res)
             redirect(%{env | url: location}, next, left - 1)
         end
```

teamon:
Handling 303 seems to be a good idea :)

I'd introduce a new private function that handles the details of preparing new request for given status instead of duplicating the redirect logic inside case match. Something simple like this should work:

```elixir
defp new_request(%{status: 303} = env, new_location), do: %{env | url: location, method: :get}
defp new_request(env, new_location), do: %{env | url: location}
```

(Bonus points for adding a reference comment explaining "why" 303 is handled differently!)

mpinkston:
Allrighty, found time to give it another shot. Simplified with preservation of the original request and support for that 303.

PR #208, opened 2018-06-12
Adds 303 as the status code that follow redirection

chollier:
303 is a valid http redirection code, see https://httpstatuses.com/303


teamon:
Thanks! 

I'm closing this as 303 was added as part of https://github.com/teamon/tesla/pull/203

PR #210, opened 2018-06-12
Use @typep for option to exclude doc generation

chevinbrown:
@teamon this makes sense to me, but I have very little context about the whole project. 🤜

### DIFF
--- a/lib/tesla/builder.ex
+++ b/lib/tesla/builder.ex
@@ -10,13 +10,13 @@ defmodule Tesla.Builder do
       Module.register_attribute(__MODULE__, :__middleware__, accumulate: true)
       Module.register_attribute(__MODULE__, :__adapter__, [])
 
-      @type option ::
-              {:method, Tesla.Env.method()}
-              | {:url, Tesla.Env.url()}
-              | {:query, Tesla.Env.query()}
-              | {:headers, Tesla.Env.headers()}
-              | {:body, Tesla.Env.body()}
-              | {:opts, Tesla.Env.opts()}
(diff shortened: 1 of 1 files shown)


chevinbrown:
@teamon I'm closing this. Doesn't feel right, unless you are totally ok and it makes sense for the options to be private.

teamon:
Hmmmmm making the type private would be a breaking change :(

chevinbrown:
Tests still pass. ;)

chevinbrown:
The root issue here is that `@typedoc false` isn't honored. 
This may be an actual elixir-lang issue, which I'll need to investigate and probably fix there...

amatalai:
I think that better solution would be to wrap current `@type` in `if unquote(docs) do`, but without `else` clause

teamon:
This would still be possible breaking change for people using `Tesla.Env.option` type in their own code (with `docs: false`)

amatalai:
hmm, but it's not `Tesla.Env.option`. It's `Foo.option` in module `Foo` with `use Tesla, docs: false` and it looks like bug to add Tesla related stuff to docs when `docs: false` is used.

Also, Elixir itself allows breaking changes when it's necessary to fix bug.
https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Compatibility%20and%20Deprecations.md#compatibility-between-elixir-non-major-versions

chevinbrown:
@amatalai @teamon seems like it may be an issue with exdoc...I'll submit a PR (#211) that should provide the correct foundation.

PR #211, opened 2018-06-13
Add @typedoc catch for docs: false

chevinbrown:
The intent of this is to provide a method to ignore `@typedoc` in shared libraries. 

Currently, ExDoc will not honor this, but this provides the footing for it to be so.


chevinbrown:
This has been fixed: https://github.com/elixir-lang/ex_doc/pull/858

PR #222, opened 2018-07-09
Add infinity retry option

Ljzn:
Sorry for so long. #201 could not be modified, so I created this. Just formatted code so that it can pass the CI.

### DIFF
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -15,7 +15,7 @@ defmodule Tesla.Middleware.Retry do
 
   ### Options
   - `:delay`        - number of milliseconds to wait before retrying (defaults to 1000)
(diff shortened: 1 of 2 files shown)


teamon:
Hmmm just by reading the code I don't think this will work - it looks like infinite recursion without ever calling `Tesla.run`

Ljzn:
@teamon  fixed

teamon:
Thanks! Can you think of some tests for this?

teamon:
@Ljzn Can you share some more details what is your specific use case for infinite retries?

teamon:
@Ljzn After giving it a lot of thought and gathering feedback from others I've decided to not include `:infinity` retry option. It is very hard to test and the sole existence of such option might encourage its usage, which is a bad practice in general. As an alternative one can always use a big number like `9999999999` as a retry value.

PR #228, opened 2018-07-16
Docs improvement

amatalai:
* Code inside ` ```ex ` was not colored properly in hexdocs
* Tesla.Migration is not public api
* Removed comment about adding `:jason` to applications. It's no longer necessary, because oldest supported elixir version is 1.4 now
* Experimental use of docs grouping feature


teamon:
ref #228

teamon:
😍 😍 😍 😍 

The only thing I'd change is to inline `docs.exs` into `mix.exs`

amatalai:
> The only thing I'd change is to inline docs.exs into mix.exs

check now

https://cl.ly/3Q0C3v1m0B2F
That looks nice, mock and multipart are no longer hidden behind all middlewares

PR #231, opened 2018-07-19
Adding check for invalid arguments in Multipart.add_field/4

### LINKED ISSUE #230: Requests hang forever when using integers as values in multipart
### Environment

`uname -a` output:
```
Linux hostname-pc 4.17.3-100.fc27.x86_64 #1 SMP Tue Jun 26 14:19:03 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```
`elixir --version` output: 
```
Erlang/OTP 20 [erts-9.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:10] [hipe] [kernel-poll:false]

Elixir 1.6.2 (compiled with OTP 19)
```

### Current behavior

When passing a integer value as third argument to `add_field/4`, it never returns when executing the request:

```elixir
alias Tesla.Multipart
chat_id_int = -1_001_143_002_998
token = Application.get_env(:tg, :token)

mp =
  Multipart.new()
  |> Multipart.add_field("chat_id", chat_id_int)
  |> Multipart.add_file("test.txt", name: "document")

Tesla.post("https://api.telegram.org/bot#{token}/sendDocument", mp) # Hangs forever
```

This works fine:

```elixir
alias Tesla.Multipart
chat_id_str = "-1001143002998"
token = Application.get_env(:tg, :token)

mp =
  Multipart.new()
  |> Multipart.add_field("chat_id", chat_id_str)
  |> Multipart.add_file("test.txt", name: "document")

Tesla.post("https://api.telegram.org/bot#{token}/sendDocument", mp)
```

### Expected behavior

I would expect it to either work fine with integers or to throw an error when I try to pass something that is not a string, whichever makes more sense.

HTiago:
Closes #230

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -3,7 +3,7 @@ if Code.ensure_loaded?(:hackney) do
     @moduledoc """
     Adapter for [hackney](https://github.com/benoitc/hackney)
 
-    Remember to add `{:hackney, "~> 1.6"}` to dependencies (and `:hackney` to applications in `mix.exs`)
+    Remember to add `{:hackney, "~> 1.13"}` to dependencies (and `:hackney` to applications in `mix.exs`)
     Also, you need to recompile tesla after adding `:hackney` dependency:
 
     ```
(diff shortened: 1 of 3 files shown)


chulkilee:
First of all, please remove merge commit :)

I'm not sure it's a good idea to use that undocumented function (`impl_for`). Also, what about impact on performance? 

I'm not sure it's idiomatic way to enforce this kind of "type check" only to raise errors. If we fix the original issue (as shown in #230) - so that it returns error somewhere, I don't think this change is needed. This should be caught by dialyzer I believe.

See the typespec:

```elixir
  @type part_stream :: IO.Stream.t() | File.Stream.t() | Enumerable.t()
  @type part_value :: String.t() | part_stream
```

teamon:
Solved by #251

PR #237, opened 2018-08-25
Add mock function into Timeout Task's process dictionary

### LINKED ISSUE #157: Tesla.Mock not compatible with Tesla.Middleware.Timeout
`Tesla.Mock` is not compatible with clients that use the `Tesla.Middleware.Timeout` middleware.

```
iex(1)> Application.put_env :tesla, :adapter, :mock
iex(2)> require Tesla
iex(3)> Tesla.Mock.mock(fn env -> %{env | status: 200} end)
iex(4)> client = Tesla.build_client([{Tesla.Middleware.Timeout, timeout: 2_000}])
iex(5)> Tesla.get(client, "/")
** (Tesla.Mock.Error) There is no mock set for process #PID<0.450.0>.
Use Tesla.Mock.mock/1 to mock HTTP requests.

See https://github.com/teamon/tesla#testing

    (tesla) lib/tesla/middleware/timeout.ex:60: Tesla.Middleware.Timeout.repass_error/1
    (tesla) lib/tesla/middleware/timeout.ex:35: Tesla.Middleware.Timeout.call/3
```

The reason is that `Tesla.Mock` looks for the mock in the current process' dictionary and the `Timeout` middleware runs the Tesla pipeline in a separate `Task` to be able to terminate it on timeout.

As a workaround, I'm going to disable the `Timeout` middleware in the test env, but this is kludgy. I'd rather have the mock support the `Timeout` middleware.

I see three ways to do this:
1. very dirty: change the implementation of the `Timeout` middleware to copy the mock into the `Task`'s process dictionary
2. dirty: change the implementation of the `Timeout` middleware to keep the parent process pid in the process dictionary; the `Tesla.Mock.pdict_get()` will use that to look for the mock function in the parent's process dictionary
3. clean: change the API of the mock spec so that it doesn't land in the process dictionary:
```
client = Tesla.Mock.mock(client, fn -> ... end)
module = Tesla.Mock.mock(MyApi, fn -> ... end)
```
The `mock` function would put the mock spec somewhere in the middlewares. Ideally, this would replace the current adapter with specced mock adapter - this has the added benefit that one no longer needs to set the `:mock` adapter in tests.

The last point would actually be great thing to do, since I have the following use case: I have a set of unit tests for my API module where I mock the adapter and verify it's behavior. But for all other tests I've built a fake implementation of the API which is started alongside the main app - this way the tests exercise the same code path as the production code. For that reason, I would strongly prefer solution 3.

Let me know which one works for you and I'll happily prepare a PR.

log4b:
Hello! This is a proposal to fix #157.

I followed @stefanchrobot first suggestion and copied the mock inside `Task` process dictionary.

On the unit tests, I had to define another client since I couldn't figure out a way to reuse the existing one's for this test. Would appreciate any feedback on how to improve it :)

### DIFF
--- a/lib/tesla/middleware/timeout.ex
+++ b/lib/tesla/middleware/timeout.ex
@@ -37,8 +37,14 @@ defmodule Tesla.Middleware.Timeout do
   end
 
   defp safe_async(func) do
+    mock_fun = Process.get(Tesla.Mock)
+
     Task.async(fn ->
       try do
+        if(is_function(mock_fun)) do
+          Process.put(Tesla.Mock, mock_fun)
+        end
+
         {:ok, func.()}
       rescue
         e in _ ->
--- a/test/tesla/middleware/timeout_test.exs
+++ b/test/tesla/middleware/timeout_test.exs
@@ -104,7 +104,7 @@ defmodule Tesla.Middleware.TimeoutTest do
 
   describe "using Tesla.Mock and timeouts" do
     test "should return the mocked response instead of mock error" do
-      Application.put_env(:tesla, :adapter, Tesla.Mock)
+      Application.put_env(:tesla, MockedClient, adapter: Tesla.Mock)
 
       Tesla.Mock.mock(fn %{method: :get, url: "https://mocked.test.com/path"} ->
         %Tesla.Env{status: 200}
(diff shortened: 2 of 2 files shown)


teamon:
Thanks for the PR!

The reason I did not implement this solution is that it is polluting production code with test-specific code. Please see https://github.com/teamon/tesla/issues/240 (point C) for the proposal on how to fix this issue.

log4b:
It makes totally sense. I'll take a closer look on #240. Closing for now. Thanks :)

Issue #240, opened 2018-08-29
1.2 draft


teamon:
## Tesla 1.2 Draft

This is an early draft of proposed changes for next version of tesla.
All feedback is more than welcome - the more I know how you all use this library the better it can become 💃 

There are three main points I'd like to tackle (A, B and C). This does not mean any other issues/improvements won't make it into the next release, these are the biggest & hardest ones. 

## A. Promote macro-less clients

Instead of 

```elixir
defmodule Github do
  use Tesla

  plug Tesla.Middleware.BaseUrl, "https://api.github.com"
  plug Tesla.Middleware.JSON

  adapter Tesla.Adapter.Hackney

  def repos, do: get("/repos")
end
```

use this one

```elixir
defmodule Github do
  @middleware [
    {Tesla.Middleware.BaseUrl, "https://api.github.com"},
    Tesla.Middleware.JSON
  ]

  @adapter Tesla.Adapter.Hackney

  def repos, do: Tesla.get(client(), "/repos")

  defp client, do: Tesla.client(@middleware, @adapter)
end
```

or with dynamic middleware/adapter

```elixir
defmodule MyApi do
  @default_middleware [
    {Tesla.Middleware.BaseUrl, "https://api.github.com"},
    Tesla.Middleware.JSON
  ]

  def items(client \\ new()), do: Tesla.get(client, "/items")

  def new(user_token, opts \\ []) do
    middleware = [{Tesla.Middleware.Headers, [{"authorization", "Bearer #{user_token}"}]}]
    Tesla.client(middleware ++ @default_middleware, opts[:adapter] || @adapter)
  end
end
```

The reasons for such change are (unordered):
- Do not worry about polluting your module with public `get/post/...`  functions
- No more name collision (I'm looking at you `delete` function)
- It becomes clear what happens with middleware arguments (#220)
- There is no difference in readability compared to the macro-based version
- It allows greater flexibility to decide on middleware/adapter at runtime (#219).

The changes required are very minimal:
- [x] Introduce `Tesla.client/1` and `Tesla.client/2` (Should it be called `Tesla.new`?)
- [x] Update adapter README with configuration examples
- [x] Update README and examples (while keeping the DSL docs!)

## ---

WDYT?

amatalai:
> Introduce Tesla.client/1 and Tesla.client/2 (Should it be called Tesla.new?)

I vote for `client`. Tesla.new would be a good name if it was supposed to create `%Tesla{}` struct

[...]

PR #243, opened 2018-09-07
Expose the timeout parameter in ibrowse

argl:
Allow passing a :timeout option to the ibrowse adapter, otherwise it will always use its hardcoded 30 second limit.

Example:
```
client =Tesla.build_client()
get(client, "/something", opts: [adapter: [timeout: 120_000]])
```


teamon:
This aligns nicely with #240, thanks!

PR #244, opened 2018-09-07
Introduce Tesla.client/1 & Tesla.client/2

teamon:
This is first part of #240

- Introduced `Tesla.client/1` & `Tesla.client/2` to configure client in runtime
- Deprecate `Tesla.build_client` & `Tesla.build_adapter`

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -91,9 +91,9 @@ defmodule Tesla.Middleware do
 
       plug Tesla.Middleware.BaseUrl, "https://example.com"
 
-  or inside tuple in case of dynamic middleware (`Tesla.build_client/2`)
+  or inside tuple in case of dynamic middleware (`Tesla.client/1`)
 
-      Tesla.build_client([{Tesla.Middleware.BaseUrl, "https://example.com"}])
+      Tesla.client([{Tesla.Middleware.BaseUrl, "https://example.com"}])
 
   ## Writing custom middleware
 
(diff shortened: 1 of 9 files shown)


lib/tesla.ex
```diff
@@ -64,13 +64,18 @@ defmodule Tesla.Env do
 end
 
 defmodule Tesla.Client do
+  @type adapter :: atom | {atom, Tesla.Env.opts()}
+  @type middleware :: atom | {atom, Tesla.Env.opts()}
```
jcabotc: Maybe `module` (instead of `atom`) is more descriptive here.

lib/tesla/builder.ex
```diff
@@ -169,7 +169,9 @@ defmodule Tesla.Builder do
     end
   end
 
-  def client(pre, post), do: %Tesla.Client{pre: runtime(pre), post: runtime(post)}
+  def client(pre, post, adapter \\ nil) do
```
jcabotc: I'd change it to something like:
```elixir
def client(pre, post) do
  %Tesla.Client{pre: runtime(pre), post: runtime(post)}
end

def client(pre, post, adapter) do
  %Tesla.Client{pre: runtime(pre), post: runtime(post), adapter: runtime(adapter)}
end
```
and remove this line: https://github.com/teamon/tesla/pull/244/files#diff-a2c274aba9c954a50446941dc12ebd88R207

Otherwise `nil` will be allowed as the list of middlewares.

lib/tesla.ex
```diff
@@ -421,26 +427,63 @@ defmodule Tesla do
...
   ```
-  defmodule ExampleAPI do
-    use Tesla
+  # add dynamic middleware
+  client = Tesla.client([{Tesla.Middleware.Headers, [{"authorization", token}]}])
+  Tesla.get(client, "/path")
+
+  # configure adapter in runtime
+  client = Tesla.client([], Tesla.Adapter.Hackney)
+  client = Tesla.client([], {Tesla.Adapter.Hackney, pool: :my_pool)
```
amatalai: missing `}`

PR #251, opened 2018-10-03
Raise ArgumentError if multipart field is invalid.

### LINKED ISSUE #230: Requests hang forever when using integers as values in multipart

ConnorRigby:
I thought i'd tackle this one since https://github.com/teamon/tesla/pull/231/ may or may not be abandoned.

Fixes #230

### DIFF
--- a/lib/tesla/multipart.ex
+++ b/lib/tesla/multipart.ex
@@ -179,7 +179,7 @@ defmodule Tesla.Multipart do
     |> Enum.join("")
   end
 
-  @spec assert_part_value(any) :: :ok | no_return
+  @spec assert_part_value!(any) :: :ok | no_return
   defp assert_part_value!(%maybe_stream{})
        when maybe_stream == IO.Stream
        when maybe_stream == File.Stream,
--- a/lib/tesla/multipart.ex
+++ b/lib/tesla/multipart.ex
@@ -34,8 +34,8 @@ defmodule Tesla.Multipart do
   @boundary_chars "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
                   |> String.split("")
 
-  @type part_stream :: IO.Stream.t() | File.Stream.t() | Enumerable.t()
-  @type part_value :: String.t() | part_stream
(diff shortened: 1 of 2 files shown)


teamon:
This looks good!

I'd add some more tests, maybe even for every allowed value.

ConnorRigby:
:+1: will take a second pass to add more tests.

ConnorRigby:
While writing the test for Enumerable, i realized that the argument isn't actually an Enumerable, but `iodata` (binary or charlist). Should the docs be updated to reflect that? A map is actually not really supported.

teamon:
Let's keep it simple and stick to the values we know work 100%

ConnorRigby:
Okay i've updated it one last time to only support `iodata` and stream data. I believe it's ready now :+1:

PR #266, opened 2018-11-22
Filter sensitive headers in debug logs with :sanitize_headers

hasclass:
Basically the same as Phoenix' `config :phoenix, :filter_parameters, ["password", "secret"]` but for Tesla debug logs and headers.

```elixir
# your client
plug Tesla.Middleware.Logger, filter_headers: ["authorization"]

# >> REQUEST >>>
# authorization: [FILTERED]
# user-agent: Mozilla
# << RESPONSE <<
# ...
```

While you generally don't want to enable debug logs in production, it can be useful to do so once in a while.

### DIFF
--- a/lib/tesla/middleware/logger.ex
+++ b/lib/tesla/middleware/logger.ex
@@ -54,7 +54,7 @@ defmodule Tesla.Middleware.Logger do
 
   ### Options
   - `:log_level` - custom function for calculating log level (see below)
-
+  - `:sanitize_headers` - sanitizes sensitive headers before logging in debug mode
 
   ## Custom log format
 
@@ -108,6 +108,16 @@ defmodule Tesla.Middleware.Logger do
   # config/dev.local.exs
   config :tesla, Tesla.Middleware.Logger, debug: false
   ```
+  #### Sanitize headers
+
+  To sanitize sensitive headers such as `authorization` in
+  debug logs, add them to the `:sanitize_headers` option.
+
+  ```
+  # config/dev.local.exs
+  config :tesla, Tesla.Middleware.Logger,
+    sanitize_headers: ["authorization"]
+  ```
   """
 
   alias Tesla.Middleware.Logger.Formatter
@@ -121,10 +131,12 @@ defmodule Tesla.Middleware.Logger do
 
   def call(env, next, opts) do
     {time, response} = :timer.tc(Tesla, :run, [env, next])
+
     level = log_level(response, opts)
     Logger.log(level, fn -> Formatter.format(env, response, time, @format) end)
 
     if Keyword.get(@config, :debug, true) do
+      env = sanitize_headers(env, Keyword.get(opts, :sanitize_headers, []))
       Logger.debug(fn -> debug(env, response) end)
     end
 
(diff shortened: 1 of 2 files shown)


test/tesla/middleware/logger_test.exs
```diff
@@ -177,6 +177,38 @@ defmodule Tesla.Middleware.LoggerTest do
...
+        {"other-header", "is not filtered"}
+      ]
+
+      log = capture_log(fn -> ClientWithFilterHeaders.get("/ok", headers: headers) end)
+
+      assert log =~ "authorization: [FILTERED]"
+      assert log =~ "other-header: is not filtered"
+    end
+  end
+
```
teamon: I know this is how other tests look like, but please use the _new_ way:

```elixir
  setup do
    # ...
    middleware = [{Tesla.Middleware.Logger, filter_headers: ["authorization"]}]
    adapter = fn env -> {:ok, %{env | status: 200, body: "ok"}} end
    client = Tesla.client(middleware, adapter)
    %{client: client}
    end

  test "sanitizes given header values", %{client: client} do
    # ...
    log = capture_log(fn -> Tesla.get(client, "/ok", headers: headers) end)
    # ...
  end
```

lib/tesla/middleware/logger.ex
```diff
@@ -54,7 +54,7 @@ defmodule Tesla.Middleware.Logger do
 
   ### Options
   - `:log_level` - custom function for calculating log level (see below)
-
+  - `:filter_headers` - sanitizes sensitive headers before logging in debug mode
```
teamon: I'm thinking about the naming: the option is named `filter...` but then in the description it uses `sanitize` - maybe the option should be just named `sanitize_headers` ?

teamon:
Nice! 👏 👏 👏

lib/tesla/middleware/logger.ex
hasclass: i had `sanitize_headers` before but decided to go with `filter_headers` because that's what phoenix and rails also call it.

lib/tesla/middleware/logger.ex
```diff
@@ -213,8 +212,17 @@ defmodule Tesla.Middleware.Logger do
     |> Enum.map(fn {k, v} -> ["Query: ", to_string(k), ": ", to_string(v), ?\n] end)
   end
 
-  defp debug_headers([]), do: @debug_no_headers
-  defp debug_headers(headers), do: Enum.map(headers, fn {k, v} -> [k, ": ", v, ?\n] end)
+  defp debug_headers([], _opts), do: @debug_no_headers
+
+  defp debug_headers(headers, opts) do
+    filtered = Keyword.get(opts, :filter_headers)
+    filtered = if is_list(filtered), do: filtered, else: []
```
teamon: I wouldn't do this as it will fail silently when you pass `filter_headers: %{"x" => true}` for example. It's better to just blow the code when something wrong is passed, so I'd keep `Keyword.get(opts, :filter_headers, [])`

hasclass: True. Let it crash.

lib/tesla/middleware/logger.ex
```diff
@@ -121,10 +131,12 @@ defmodule Tesla.Middleware.Logger do
 
   def call(env, next, opts) do
     {time, response} = :timer.tc(Tesla, :run, [env, next])
+
```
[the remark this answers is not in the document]
hasclass: that empty new line I introduced isn't picked up by `mix format`. I removed it manually now.

hasclass:
Fantastic! Thanks for your time and support @teamon

PR #267, opened 2018-12-05
Allow nil path on BaseUrl middleware to use the base as-is

chulkilee:
To reuse client, I use `Tesla.Middleware.BaseUrl` with full url (i.e. with path). Currently the middleware does not provide a way to use the base url as-is.

This change introduces new behaviour: this was not allowed previously (it gives an error to concat nil to string), so it's not breaking change.

### DIFF
--- a/lib/tesla/middleware/core.ex
+++ b/lib/tesla/middleware/core.ex
@@ -15,6 +15,7 @@ defmodule Tesla.Middleware.BaseUrl do
     plug Tesla.Middleware.BaseUrl, "https://api.github.com"
   end
 
+  MyClient.get(nil) # equals to GET http://example.com
   MyClient.get("/path") # equals to GET https://api.github.com/path
   MyClient.get("http://example.com/path") # equals to GET http://example.com/path
   ```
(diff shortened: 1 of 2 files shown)


amatalai:
It is possible to pass empty string

```elixir
defmodule Foo do
  use Tesla
  plug Tesla.Middleware.BaseUrl, "http://localhost:4000/ping"
end

iex> Foo.get("")
{:ok, %Tesla.Env{body: "pong", status: 200, url: "http://localhost:4000/ping/"}}
```

chulkilee:
`"http://localhost:4000/ping"` is not same with "http://localhost:4000/ping/" :)

chulkilee:
Since `Tesla.Env.url()` type is `binary` so this would introduce dialyzer error. I do not want to add make that nullable :( hm..

chulkilee:
BTW I now use following simple middleware since I do not override any URL when using clients with it.

```elixir
  defmodule BaseUrlMiddleware do
    @moduledoc false

    @behaviour Tesla.Middleware
    def call(env, next, base), do: Tesla.run(%{env | url: base}, next)
  end
```

teamon:
Btw, we could probably skip the regex here:

```ex
defp apply_base(env, base) do
  case env.url do
    nil -> env
    "https://" <> _ -> env
    "http://" <> _ -> env
    %{env | url: join(base, env.url)}
  end
end
```

(not tested)

chulkilee:
I don't think adding nil to url type is not a good option - and instead it's better to keep the original url when empty string is given. See #268

PR #268, opened 2018-12-07
Base url improve: case-insensitive scheme and keep empty path

chulkilee:
Replace #267

## Changes

- Update doc and test for base url with path
- Previously, when empty url is given, the result url would have always `/` at the end. Now this behavior is changed to keep the original url (as in the base url middleware) as-is.
- Allow different case for http(s).

## Note

- No change for type (`url` is still expected be a string)
- I thought of using `URI.parse/1`, but just leave the current behaviour (passing only http(s))

```elixir
  defp apply_base(%{url: url} = env, base) do
    case URI.parse(url) do
      %{scheme: nil} -> %{env | url: join(base, url)}
      _ -> env
    end
  end
```

### DIFF
--- a/lib/tesla/middleware/core.ex
+++ b/lib/tesla/middleware/core.ex
@@ -27,7 +27,7 @@ defmodule Tesla.Middleware.BaseUrl do
   end
 
   defp apply_base(env, base) do
-    if Regex.match?(~r/^https?:\/\//, env.url) do
+    if Regex.match?(~r/^https?:\/\//i, env.url) do
       # skip if url is already with scheme
       env
     else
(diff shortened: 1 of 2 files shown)


chulkilee:
Great summary. I'm more inclined to consider it as a bug fix since the doc does not specify that behavior.

I guess we have three options

- change the behavior, and bump the patch version (as a bug fix)
- change the behavior, and bump the major or minor version (as a behavior change )
- introduce new middleware (different name) with "correct" behavior, and deprecate this middleware (not sure `@deprecated` would work in middleware)

teamon:
I've decided to treat this as a bug fix, it will be included in the next minor release. 
Thanks @chulkilee!

stevenferrer:
While waiting for the next release, I copied the change from @chulkilee and created a custom middleware. Great work @chulkilee!

teamon:
Just FYI, v1.3.0 has been released https://github.com/teamon/tesla/releases/tag/v1.3.0

Issue #249, opened 2018-09-28
Add UrlBuilder to tesla core


teamon:
Recently I have been using this little middleware quite a lot.
The reasoning for such way of generating paths is monitoring - with url template + params we can report path template as metrics dimension (to [prometheus](https://github.com/teamon/tesla/wiki/Cookbook%3A-Prometheus-integration) for example).

```elixir
  defmodule UrlBuilder do
    @behaviour Tesla.Middleware

    @impl true
    def call(env, next, _) do
      url = build_url(env.url, env.opts[:params])
      Tesla.run(%{env | url: url}, next)
    end

    defp build_url(url, nil), do: url

    defp build_url(url, params) do
      Enum.reduce(params, url, fn {k, v}, u -> String.replace(u, ":#{k}", to_string(v)) end)
    end
  end
```

Can be used like:

```elixir
Tesla.get(client, "/posts/:id", opts: [params: [id: id]])
```

Questions yet to be answered:
- Does it makes sense to make it built-in into tesla?
- If yes, I believe the naming could be improved (both middleware name and vague `params` opt)

WDYT @amatalai @hodak @ericmj @jcabotc ?

chulkilee:
What about having separate `tesla-middleware-contrib`?

teamon:
`Tesla.Middleware.Instrumentation` seems like a good idea, it could have then configurable "adapter to middleware" for prometehus/statsd/etc.

+1 for `PathParams ` + `opts: [path_params: [id: 123]]`

teamon:
Instrumentation should be done with #260.
Regarding `PathParams` - I think it can be added to the core :)

PR #270, opened 2018-12-17
PathParams middleware

teamon:
Implementation of #249 

TODO:
- [x] Update README
- [x] Add module docs

cc @jcabotc


chulkilee:
Following cases should be considered and the expected behaviour must be in the test.

- when value is not given: raise or pass the original value?
- when nil value is given: should it replace it with empty string or `nil` or `null` or raise error?
- when value with special value is given, such as `/`, `#`, `?`: should it be urlencoded or not?
- when a placeholder contains another placeholder as prefix: e.g. `:id`, `:id_new`
- when a placeholder is part of query string or value, not path (e.g. `/users?count=:count`)

teamon:
@chulkilee <3 

I'd say, that:
- when value is not given: keep placeholder `"/users/:x" & [y: 1] ==> "/users/:id"`
- when nil value is given: I'm leaning towards requiring a binary, with auto `to_string()` as a second option
- when value with special value is given, such as /, #, ?: should it be urlencoded or not?: I think it should
- when a placeholder contains another placeholder as prefix: e.g. :id, :id_new: Ouh, nice catch. We *could* sort by key length or use regex to extract `:[\w_]+` and do exact match
- when a placeholder is part of query string or value, not path (e.g. /users?count=:count): I wouldn't support not forbid it - there is `query: [a: 1]` for this case

PR #278, opened 2019-01-14
Add support for Hypertext Application Language

### LINKED ISSUE #277: Add Support for Hypertext Application Language
HAL (http://stateless.co/hal_specification.html) is a simple format that gives a consistent and easy way to hyperlink between resources in your API.

ordepdev:
HAL (http://stateless.co/hal_specification.html) is a simple format that gives a consistent and easy way to hyperlink between resources in your API.

The goal of this PR is to add this _content type_ to the default ones.

Closes #277.

### DIFF
--- a/lib/tesla/middleware/json.ex
+++ b/lib/tesla/middleware/json.ex
@@ -41,7 +41,7 @@ defmodule Tesla.Middleware.JSON do
 
   # NOTE: text/javascript added to support Facebook Graph API.
   #       see https://github.com/teamon/tesla/pull/13
-  @default_content_types ["application/json", "text/javascript"]
+  @default_content_types ["application/json", "application/hal+json", "text/javascript"]
   @default_encode_content_type "application/json"
   @default_engine Jason
 
--- a/test/tesla/middleware/json_test.exs
+++ b/test/tesla/middleware/json_test.exs
@@ -42,6 +42,10 @@ defmodule Tesla.Middleware.JsonTest do
             "/stream" ->
               list = env.body |> Enum.to_list() |> Enum.join("---")
               {200, [], list}
+
+            "/haljson" ->
+              {200, [{"content-type", "application/hal+json"}],
+               "{\"id\": 123, \"_links\":{\"self\":{\"href\":\"https://foo.bar/123\"}}}"}
           end
 
         {:ok, %{env | status: status, headers: headers, body: body}}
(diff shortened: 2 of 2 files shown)


teamon:
This seems like a specific case that can be easily handled by configuration without the need to change tesla. 

Adding support for text/javascript was a mistake (not all text/javascript content is valid json) and I don’t want to make the same one again. 

Having said that, can you share some more context on why have you decided to submit this PR? Maybe there is something that should be done to make it more clear.

ordepdev:
Hey @teamon 👋

> This seems like a specific case that can be easily handled by configuration without the need to change tesla.

Indeed, I'm actually using `plug(Tesla.Middleware.JSON, decode_content_types: ["application/hal+json"])` and it works like a charm 🙌

> Having said that, can you share some more context on why have you decided to submit this PR? Maybe there is something that should be done to make it more clear.

HAL is a generic media type with which Web APIs can be developed and exposed as a series of links.  Clients of these APIs can select links by their link relation type and traverse them in order to progress through the application.

HAL's conventions result in a uniform interface for serving and consuming hypermedia, enabling the creation of general-purpose libraries that can be re-used on any API utilizing HAL.

The primary design goals of HAL are generality and simplicity.  HAL can be applied to many different domains and imposes the minimal amount of structure necessary to cover the key requirements of a hypermedia Web API.

More and more we see RESTful APIs with this type of response:

```
   GET /orders HTTP/1.1
   Host: example.org
   Accept: application/hal+json

   HTTP/1.1 200 OK
   Content-Type: application/hal+json

   {
     "_links": {
       "self": { "href": "/orders" },
       "next": { "href": "/orders?page=2" },
       "find": { "href": "/orders{?id}", "templated": true }
     },
     "_embedded": {
       "orders": [{
           "_links": {
             "self": { "href": "/orders/123" },
             "basket": { "href": "/baskets/98712" },
             "customer": { "href": "/customers/7809" }
           },
           "total": 30.00,
           "currency": "USD",
           "status": "shipped",
         },{
           "_links": {
             "self": { "href": "/orders/124" },
             "basket": { "href": "/baskets/97213" },
             "customer": { "href": "/customers/12369" }
           },
           "total": 20.00,
           "currency": "USD",
           "status": "processing"
       }]
     },
     "currentlyProcessing": 14,
     "shippedToday": 20
   }
```

This architectural style lets you use hypermedia links in the response contents so that the client can dynamically navigate to the appropriate resource by traversing the hypermedia links. This is conceptually the same as a web user navigating through web pages by clicking the appropriate hyperlinks in order to achieve a final goal.

> Adding support for text/javascript was a mistake (not all text/javascript content is valid json) and I don’t want to make the same one again.

That's why I'm suggesting to add `application/hal+json` to the defaults. It should be _natural_ and _transparent_ to use this format whithout declaring it explicitly.

Tell me what you think and thanks for taking your time 🤗

teamon:
@ordepdev I get what you're saying, yet I stand at my position that it should not be part of the core.It can be easily done with the current tesla.

Again, adding `"text/javascript"` was a mistake in the early days, it might get removed in 2.0 but so far I'm not breaking backwards compatibility for aesthetic reasons only.

@ordepdev I appreciate your contribution and the time you've spend to describe the case. Unfortunately I have to say no, I'm sure you understand that. Thank you again, and I hope this won't put you away from future contributions.

PR #280, opened 2019-01-24
Added elixir 1.8 to travis matrix and fixed new deprecation warnings

amatalai:


lib/tesla.ex
```diff
@@ -475,7 +475,7 @@ defmodule Tesla do
   MyApi.get_something(client, 42)
   ```
   """
-  @since "1.2.0"
+  if Version.match?(System.version(), "~> 1.7"), do: @doc(since: "1.2.0")
```
teamon: Shouldn’t there be `, else: @since "1.2.0"`?

amatalai: It could be but it will cause warnings in some older elixir version which we still support

Alternatively I though about something like
```elixir
defmodule Tesla.HandleDeprecations do
  @moduledoc false
  
  defmacro since(version) do
    cond do
      Version.match?(System.version(), "~> 1.7") ->
        quote do: @doc since: unquote(version)

      Version.match?(System.version(), "~> 1.5.3") ->
        quote do: @since unquote(version)

      true ->
        :nothing
    end
  end
end
```
and then simple `since("1.2.0")` in regular code. But I wasn't sure if this won't be overkill as documentation for hexdocs can be generated with elixir version that supports `@doc since:`
wdyt?

PR #284, opened 2019-03-13
Add Clack middleware that adds X-Clacks-Overhead header

hauleth:
This is small easter egg that's purpose is to pay tribute to Sir Terry
Pratchett and to keep his memory alive forever (or at least as long
internet will stand).


hodak:
It would be cool in a separate repository for those who are into those kinds of things, but I don't think it should be merged into core tesla

teamon:
I agree with @hodak, this can be a separate package.

PR #289, opened 2019-03-29
Fix the ci building

RyanSiu1995:
I have added the function pattern matching for some failed builds. Telemetry will enclose the value given with a map after processing. 
Hope this pull request can fix those build failure and unblock the development of Tesla :)

### DIFF
--- a/test/tesla/middleware/telemetry_test.exs
+++ b/test/tesla/middleware/telemetry_test.exs
@@ -23,8 +23,12 @@ defmodule Tesla.Middleware.TelemetryTest do
     :telemetry.attach(
       "telemetry_test",
       [:tesla, :request],
-      fn [:tesla, :request], response, meta, _config ->
-        send(self(), {:ok_called, is_integer(response), meta})
+      fn
+        [:tesla, :request], %{value: time}, meta, _config ->
+          send(self(), {:error_called, is_integer(time), meta})
+
+        [:tesla, :request], response, meta, _config ->
+          send(self(), {:ok_called, is_integer(response), meta})
       end,
       nil
     )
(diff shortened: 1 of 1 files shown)


test/tesla/middleware/telemetry_test.exs
```diff
@@ -23,8 +23,12 @@ defmodule Tesla.Middleware.TelemetryTest do
     :telemetry.attach(
       "telemetry_test",
       [:tesla, :request],
-      fn [:tesla, :request], response, meta, _config ->
-        send(self(), {:ok_called, is_integer(response), meta})
+      fn
+        [:tesla, :request], %{value: time}, meta, _config ->
+          send(self(), {:ok_called, is_integer(time), meta})
+
+        [:tesla, :request], response, meta, _config ->
```
jalcine: Why not add a guard matcher here to check if `response` is an integer?

teamon:
@RyanSiu1995  is this something that changed in recent telemetry versions?

RyanSiu1995:
Well... I don't know as I didn't keep in touch with the development in telemetry. But even if there is an update from telemetry, the mix.lock make sure our telemetry version is exact version as I developed with. So I don't know the root cause making the failure... Sorry bro :(

test/tesla/middleware/telemetry_test.exs
RyanSiu1995: Yeah, the guard is more concise and precise here. 
I want to know if it is proper to raise an assertion error from guard/pattern match in elixir for unit test. As I come from Java, most of my unit tests I have written indicate the assertion in ```assertThat()```. 
Anyway thank you for the comment

jalcine: Totally, no problem! Thank _you_ for working on this!

teamon:
@RyanSiu1995 Can you link the failed builds? Adding another, less restrictive pattern seems weird and hackish.

RyanSiu1995:
I have made a commit to print out the telemetry output with the unit test.
You can see #10.17 has a different output as others do.
Please check the link [here](https://travis-ci.org/RyanSiu1995/tesla/builds/514115993?utm_source=github_status&utm_medium=notification).

teamon:
This is the one running on `Elixir: 1.5.3 OTP Release: 20.2` with `LOCKFILE=poison3`. I'd update the poison3 lockfile first and try to discover reasons for this message differences.

RyanSiu1995:
Finally, I got the reason. Telemetry 0.4.0 has depreciated the single number as measurement. 
In the failed test, as the lockfile has regenerated so it gets the latest release, which will change the measurement into a map. It seems 0.4.0 is not backward compatible so we can this error. 
I will make another pull request in updating the telemetry into 0.4.0 for fixing this bugs. Should we make an issue to telemetry group to alert this problem?

RyanSiu1995:
Continued on #292

PR #290, opened 2019-04-08
Use correct SSL options for Hackney

timbuchwaldt:
Setting `ssl` has no effect, the option in hackney is called `ssl_options`


teamon:
Nice catch!

PR #292, opened 2019-04-09
Update telemetry to 0.4.0 and change the API for telemetry

RyanSiu1995:
This is the proper fix on pull request #289.

### DIFF
--- a/lib/tesla/middleware/telemetry.ex
+++ b/lib/tesla/middleware/telemetry.ex
@@ -24,7 +24,7 @@ defmodule Tesla.Middleware.Telemetry do
   @doc false
   def call(env, next, _opts) do
     {time, res} = :timer.tc(Tesla, :run, [env, next])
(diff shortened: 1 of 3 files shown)


lib/tesla/middleware/telemetry.ex
```diff
@@ -24,7 +24,7 @@ defmodule Tesla.Middleware.Telemetry do
   @doc false
   def call(env, next, _opts) do
     {time, res} = :timer.tc(Tesla, :run, [env, next])
-    :telemetry.execute([:tesla, :request], time, %{result: res})
+    :telemetry.execute([:tesla, :request], %{request_time: time}, %{result: res})
```
chulkilee: Does it work with both telemetry 0.3 and 0.4? If not, probably better to bump the dep from `mix.exs`.

RyanSiu1995: It works for both telemetry.

PR #275, opened 2019-01-06
Retry middleware support custom determination function

RyanSiu1995:
Regarding the issue #263.

PR #294, opened 2019-04-24
Add exponential backoff to retry middleware

keathley:
Hey! Thanks for this great library. We've been converting a lot of our http clients to Tesla and have been really happy with it so far.

One of the things that we found ourself needing was a retry middleware that did exponential backoff with jitter. I thought I would submit what we're using as a PR to see if its interesting to y'all. I wasn't sure if I should leave the existing retry middleware as is or if I should create a new `RetryWithBackoff` middleware. I opted to go this route because it seems like in most use cases you're going to want some sort of backoff and jitter as opposed to a set delay. That said this is probably a breaking change for _someone_ because they're probably relying on the existing algorithm.

Let me know what you think. Thanks again for all of the work on Tesla :+1:.

### DIFF
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -2,9 +2,24 @@ defmodule Tesla.Middleware.Retry do
   @behaviour Tesla.Middleware
 
   @moduledoc """
-  Retry the HTTP call in case of connection error by default (`nxdomain`, `connrefused` etc).
-  Application error checking for retry can be customized through `:should_retry` option by
+  Retry using exponential backoff and full jitter. This middleware only retries in the
+  case of connection errors (`nxdomain`, `connrefused` etc). Application error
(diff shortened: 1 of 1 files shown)


teamon:
NICE!

It definitely make sense to have an exponential backoff, but we need to figure out a way to add this as an option to not break existing behaviour.

Btw, this PR will conflict with https://github.com/teamon/tesla/pull/275 which should be merged first, but it should be straightforward to resolve.

keathley:
Sounds good. I saw the other PRs and figured there would be some conflicts to work out. Not a problem to get them resolved on my end. I've been thinking about ways we could make this a non-breaking change and I can't come up with a great idea. Happy to do what you think is best :+1:.

teamon:
Hey @keathley, would you mind rebasing this PR on master so we can merge it? :)

keathley:
@teamon Rebased. Let me know if anything doesn't look right.

PR #299, opened 2019-05-16
Allows put_headers to work with a nil header argument

### LINKED ISSUE #298: No matching clause error when using DigestAuth
The DigestAuth is sending `nil` into `Tesla.put_headers/2`. I know that I can get it to work by forcing `Tesla.put_headers/2` to convert nils to an empty list, but I wasn't sure if that was the best way or if it should be taken care of higher in the stack.

~~~elixir
** (FunctionClauseError) no function clause matching in Tesla.put_headers/2

    The following arguments were given to Tesla.put_headers/2:

        # 1
        %Tesla.Env{
          __client__: %Tesla.Client{adapter: nil, fun: nil, post: [], pre: []},
          __module__: Axis,
          body: nil,
          headers: [],
          method: :get,
          opts: [digest_auth_handshake: true],
          query: [],
          status: nil,
          url: "http://192.168.1.189/some_path.cgi"
        }

        # 2
        nil

    Attempted function clauses (showing 1 out of 1):

        def put_headers(%Tesla.Env{} = env, list) when is_list(list)

    (tesla) lib/tesla.ex:416: Tesla.put_headers/2
    (tesla) lib/tesla/middleware/core.ex:66: Tesla.Middleware.Headers.call/3
    (tesla) lib/tesla/middleware/digest_auth.ex:57: Tesla.Middleware.DigestAuth.authorization_vars/2
    (tesla) lib/tesla/middleware/digest_auth.ex:47: Tesla.Middleware.DigestAuth.authorization_header/2
    (tesla) lib/tesla/middleware/digest_auth.ex:38: Tesla.Middleware.DigestAuth.call/3
~~~

adkron:
I don't know if this is the best way to fix it or not, but I thought I
would put something out there to get the conversation going. I don't
like having `nil` passed in place of an empty list, but that was already
happening above.

Amos King @adkron <amos@binarynoggin.com>

closed #298

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -412,7 +412,9 @@ defmodule Tesla do
     %{env | headers: headers}
   end
 
-  @spec put_headers(Env.t(), [{binary, binary}]) :: Env.t()
+  @spec put_headers(Env.t(), [{binary, binary}] | nil) :: Env.t()
+  def put_headers(%Env{} = env, nil), do: env
+
   def put_headers(%Env{} = env, list) when is_list(list) do
     %{env | headers: env.headers ++ list}
   end
--- a/test/tesla_test.exs
+++ b/test/tesla_test.exs
@@ -256,6 +256,10 @@ defmodule TeslaTest do
       assert get_header(env, "server") == "Cowboy"
       assert get_header(env, "content-length") == "100"
       assert get_header(env, "content-type") == "text/plain"
+
+      headers = Map.get(env, :headers)
+      env = Tesla.put_headers(env, nil)
+      assert headers == Map.get(env, :headers)
     end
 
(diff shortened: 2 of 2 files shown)


chulkilee:
DigestAuth middleware should not pass nil header. Why don't we fix DigestAuth to pass empty header (or skipping adding header)... or raise error (if that should not happen), instead of making typespec changes in `put_headers`?

adkron:
That is even better. I have no computer and I’m heading into the desert for a few days. I can dig in when I get back if someone else doesn’t get to it first. Have a wonderful day.

adkron:
I have no idea why, but this seems to be working now. I will reopen if I see this happening again.

PR #303, opened 2019-06-17
Allows DigestAuth to work with Hackney using the async option

### LINKED ISSUE #302: Support async requests using Digest Auth
I need to stream mjpg, and the only want to make that work is to have support for async. If we utilize the ability of hackney to make async requests, we would be able to support a lot of other use cases. I was trying to think of an elegant way to deal with the callbacks that would need to be in place. One thought I had is when passing async to the adapter that we make the value a function, or module, or MFA. The only strange thing with that is that it isn't a direct configuration of Hackney that we would be sending. The other idea that I had was to have a completely separate call for asynchronous requests.

adkron:
The Hackney handler was not handling the async option correctly. The problem is
that Hackney would return `{:ok, ref}` as an asynchronous response and the
caller is expected to process messages using that reference.

The Hackney adapter did not have a handler for the `{:ok, ref}`. I added a
handler for the `{:ok, ref}` response that waits for the header and status
messages from Hackney, but passes back the `ref` as the body of the response.
The reason is that some async requests are long-living and require the end user
to implement the streaming data, i.e. mjpg. The header and the status should not
take much time and are expected in many instances.

No changes were needed in the DigestAuth module since DigestAuth doesn't use the
body.

I was able to use this on an mjpg async call and use a handler for the body
messages. The handler did have to receive messages like `{:hackney_response,
ref, bin}`, but I didn't think that we should capture that message to
change the message key to something like `{:tesla_response, ref, bin}`.

closes #302 

Amos King @adkron <amos@binarynoggin.com>

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -41,7 +41,7 @@ if Code.ensure_loaded?(:hackney) do
     end
 
     defp format_body(data) when is_list(data), do: IO.iodata_to_binary(data)
-    defp format_body(data) when is_binary(data), do: data
+    defp format_body(data) when is_binary(data) or is_reference(data), do: data
 
     defp request(env, opts) do
       request(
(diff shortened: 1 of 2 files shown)


adkron:
The failing build looks like Erlang couldn't be pulled.

amatalai:
I came across the same issue in another repo  (I have already reported this on the Travis forum). I will restart the build as soon as they fix it.

amatalai:
Travis is moving their default dist to `xenial` which doesn't support `otp-18`. We have to either remove `otp-18` from matrix or explicitly set dist
```yml
    - elixir: 1.5.3
      otp_release: 18.3
      dist: trusty
```
cc @teamon 
link to travis forum for reference: https://travis-ci.community/t/erlang-18-3-couldnt-be-pulled/3836/4

teamon:
Lets set the dist for otp 18 then until tesla 2.0, nice find 👏

PR #304, opened 2019-06-19
Gun adapter

alex-strizhakov:
Adapter for gun - https://github.com/ninenines/gun

Added custom option - max_body.


lib/tesla/adapter/gun.ex
```diff
@@ -0,0 +1,209 @@
...
+      stream
+    end
+
+    defp open_stream(pid, method, url, headers, body, false),
+      do: :gun.request(pid, method, url, headers, body)
+
+    defp read_response(pid, stream, opts) do
+      receive do
+        msg ->
+          case msg do
```
rinpatch: Can't you just write clauses in the `recieve` block directly?

lib/tesla/adapter/gun.ex
```diff
@@ -0,0 +1,209 @@
...
+          {:error, "read response timeout"}
+      end
+    end
+
+    defp read_body(pid, stream, mref, opts, acc \\ "") do
+      limit = opts[:max_body]
+
+      receive do
+        msg ->
+          case msg do
```
rinpatch: Can't you just write clauses in the recieve block directly?

lib/tesla/adapter/gun.ex
```diff
@@ -0,0 +1,209 @@
...
+                read_body(pid, stream, mref, opts, acc <> part)
+              else
+                {:error, "body too large"}
+              end
+
+            msg ->
+              IO.inspect(msg)
+              IO.inspect("unsupported message received in read body.")
+              msg
+          end
```
rinpatch: Should handle [:gun_error](https://ninenines.eu/docs/en/gun/1.3/guide/http/#_processing_responses)

lib/tesla/adapter/gun.ex
```diff
@@ -0,0 +1,200 @@
...
+    mix deps.compile tesla
+    ```
+    ### Example usage
+    ```
+    # set globally in config/config.exs
+    config :tesla, :adapter, Pleroma.Tesla.Adapter.Gun
+    # set per module
+    defmodule MyClient do
+      use Tesla
+      adapter Pleroma.Tesla.Adapter.Gun
```
teamon:
```suggestion
      adapter Tesla.Adapter.Gun
```

lib/tesla/adapter/gun.ex
```diff
@@ -0,0 +1,200 @@
...
+    Remember to add `{:gun, "~> 1.3"}` to dependencies
+    Also, you need to recompile tesla after adding `:gun` dependency:
+    ```
+    mix deps.clean tesla
+    mix deps.compile tesla
+    ```
+    ### Example usage
+    ```
+    # set globally in config/config.exs
+    config :tesla, :adapter, Pleroma.Tesla.Adapter.Gun
```
teamon:
```suggestion
    config :tesla, :adapter, Tesla.Adapter.Gun
```

lib/tesla/adapter/gun.ex
```diff
@@ -0,0 +1,200 @@
...
+    end
+    ```
+
+    ### Options:
+
+    * `connect_timeout` - Connection timeout.
+    * `http_opts` - Options specific to the HTTP protocol.
+    * `http2_opts` -  Options specific to the HTTP/2 protocol.
+    * `protocols` - Ordered list of preferred protocols. Defaults: [http2, http] - for :tls, [http] - for :tcp.
+    * `retry` - Number of times Gun will try to reconnect on failure before giving up. Default: 5
```
teamon: There is `Tesla.Middleware.Retry` middleware that works with all adapters, exposing this option might be confusing (which one should be used?)

mix.exs
```diff
@@ -69,6 +70,7 @@ defmodule Tesla.Mixfile do
       # other
       {:fuse, "~> 2.4", optional: true},
       {:telemetry, "~> 0.3", optional: true},
+      {:cowlib, "~> 1.0.2", override: true, optional: true},
```
teamon: Why is this needed?

test/tesla/adapter/gun_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  use Tesla.AdapterCase.SSL
+
+  test "timeout option" do
+    request = %Env{
+      method: :get,
+      url: "#{@http}/delay/2"
+    }
+
+    assert {:error, "read response timeout"} = Tesla.Adapter.Gun.call(request, timeout: 1_000)
+  end
```
teamon: While there are some issues with Tesla.Middleware.Timeout I don't think we should expose such custom options as part of adapter interface. In general adapter options should be used as last-resort case when one absolutely have to pass some options to the underlying adapter.

mix.exs
alex-strizhakov: gun requires {:cowlib, "~> 2.6.0"}, that's why `override: true` was added

lib/tesla/adapter/gun.ex
alex-strizhakov: removed these options

test/tesla/adapter/gun_test.exs
alex-strizhakov: now it uses `opts[:timeout]` option or adapter default fallback setting

mix.exs
alex-strizhakov: removed after update `httparrot` dependency

rinpatch:
@teamon could you give this another review please?

teamon:
Please rebase to recent master (it includes CI fixes). Once all the tests pass I'd be happy to merge this MR 🚀

alex-strizhakov:
@teamon can elixir 1.4.5 and otp 18 be removed?

teamon:
I’m currently on holidays, I’ll get back to you next week.

teamon:
Dropping support for elixir 1.4 & otp 18 would be a breaking change and would require releasing tesla 2.0. 

However, if there is no easy way to provide 1.4 support I'm ok with that, 1.4 caused few other issues already and we have elixir 1.19 already.

alex-strizhakov:
[httparrot 1.2.0](https://github.com/edgurgel/httparrot/blob/master/mix.exs) requires `con_cache` 0.13, which can't be compiled on 1.4.5.

teamon:
@alex-strizhakov And do we have to bump httparrot? :)

alex-strizhakov:
@teamon without httparrot we need to have `{:cowlib, "~> 1.0.2", override: true, optional: true}` in `mix.exs`.

teamon:
... and `gun` requires `cowlib`. 

Ok, let's drop 1.4 & 18 🔥

alex-strizhakov:
> ... and gun requires cowlib.
Forgot about it 😅

PR #308, opened 2019-07-18
Add status to Tesla.option type

tylerbarker:
## Adding `status` to `Tesla.option/0`
While trying to silence some dialyzer errors I noticed for every time I was calling an equivalent of `Mock.json(json(%{"some" => "data"}, status: 404)` like in the documentation for `json/2`, ElixirLS picked up that `status` does not exist in `Tesla.option()`.

All other fields in `Tesla.t()` are included in the option type but that one, so I'm assuming it was just missed?

### DIFF
--- a/lib/tesla/builder.ex
+++ b/lib/tesla/builder.ex
@@ -16,6 +16,7 @@ defmodule Tesla.Builder do
               | {:query, Tesla.Env.query()}
               | {:headers, Tesla.Env.headers()}
               | {:body, Tesla.Env.body()}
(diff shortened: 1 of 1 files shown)


teamon:
I’m currently on holidays, I’ll get back to you next week.

teamon:
`Tesla.option` type is used to type arguments to `request/get/post/etc` functions. As this functions are for performing requests they must not contain `:status` argument. 

If anything, the documentation for `json` function should be updated.

PR #310, opened 2019-08-11
Expanding functionality of the gun adapter

alex-strizhakov:
Added possibility to get response body as stream and in chunks.
Also added possibility for reusing opened connection.

### DIFF
--- a/lib/tesla/adapter/gun.ex
+++ b/lib/tesla/adapter/gun.ex
@@ -139,19 +139,14 @@ if Code.ensure_loaded?(:gun) do
           {:ok, status, headers, ""}
 
         {:gun_response, ^pid, ^stream, :nofin, status, headers} ->
-          if opts[:stream_response] do
-            {:ok, status, headers, %{pid: pid, stream: stream}}
-          else
-            case read_body(pid, stream, opts) do
-              {:ok, body} ->
-                :gun.close(pid)
-                {:ok, status, headers, body}
-
-              {:error, error} ->
-                :gun.close(pid)
-                {:error, error}
+          body_as =
+            cond do
+              opts[:stream_response] -> :stream
+              opts[:chunks_response] -> :chunks
+              true -> :plain
             end
-          end
+
+          format_response(pid, stream, opts, status, headers, body_as)
 
         {:error, error} ->
           :gun.close(pid)
@@ -177,6 +172,42 @@ if Code.ensure_loaded?(:gun) do
       end
     end
 
+    defp format_response(pid, stream, opts, status, headers, :plain) do
+      case read_body(pid, stream, opts) do
+        {:ok, body} ->
+          :gun.close(pid)
+          {:ok, status, headers, body}
+
+        {:error, error} ->
+          :gun.close(pid)
+          {:error, error}
+      end
+    end
+
+    defp format_response(pid, stream, opts, status, headers, :stream) do
+      stream_body =
+        Stream.resource(
+          fn -> %{pid: pid, stream: stream} end,
+          fn
+            %{pid: pid, stream: stream} ->
+              case read_chunk(pid, stream, opts) do
+                {:nofin, part} -> {[part], %{pid: pid, stream: stream}}
+                {:fin, body} -> {[body], %{pid: pid, final: :fin}}
+              end
+
+            %{pid: pid, final: :fin} ->
+              {:halt, %{pid: pid}}
+          end,
+          fn %{pid: pid} -> :gun.close(pid) end
+        )
+
+      {:ok, status, headers, stream_body}
+    end
+
+    defp format_response(pid, stream, opts, status, headers, :chunks) do
+      {:ok, status, headers, %{pid: pid, stream: stream, opts: opts}}
+    end
+
     def read_chunk(pid, stream, opts) do
       receive do
         {:gun_data, ^pid, ^stream, :fin, body} ->
@@ -188,6 +219,10 @@ if Code.ensure_loaded?(:gun) do
         opts[:timeout] || @adapter_default_timeout ->
           {:error, :timeout}
       end
+    end
+
+    def close(%{pid: pid}) do
+      :gun.close(pid)
     end
 
     defp read_body(pid, stream, opts, acc \\ "") do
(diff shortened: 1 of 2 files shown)


lib/tesla/adapter/gun.ex
```diff
@@ -118,7 +134,12 @@ if Code.ensure_loaded?(:gun) do
     defp open_conn(url, opts) do
       uri = URI.parse(url)
       opts = if uri.scheme == "https", do: Map.put(opts, :transport, :tls), else: opts
-      {:ok, pid} = :gun.open(to_charlist(uri.host), uri.port, Map.take(opts, @gun_keys))
+
+      {:ok, pid} =
+        if opts[:conn],
+          do: {:ok, opts[:conn]},
+          else: :gun.open(to_charlist(uri.host), uri.port, Map.take(opts, @gun_keys))
+
```
teamon: I'd go with standard multiline `if` ;)

test/tesla/adapter/gun_test.exs
```diff
@@ -6,26 +6,27 @@ defmodule Tesla.Adapter.GunTest do
   use Tesla.AdapterCase.Multipart
   use Tesla.AdapterCase.StreamRequestBody
   use Tesla.AdapterCase.SSL
+  alias Tesla.Adapter.Gun
```
teamon: Seems to be used in single place only, might not be necessary

test/tesla/adapter/gun_test.exs
```diff
@@ -41,10 +42,114 @@ defmodule Tesla.Adapter.GunTest do
...
+
+  defp read_body(pid, stream, acc \\ "") do
+    case Tesla.Adapter.Gun.read_chunk(pid, stream, timeout: 1_000) do
+      {:fin, body} ->
+        :ok = Gun.close(pid)
+        acc <> body
+
+      {:nofin, part} ->
+        read_body(pid, stream, acc <> part)
+    end
```
teamon: It would be nice to add "reuse connection" test

lib/tesla/adapter/gun.ex
```diff
@@ -101,71 +120,141 @@ if Code.ensure_loaded?(:gun) do
...
 
-    defp request_stream(method, url, headers, body, opts) do
-      with {pid, f_url} <- open_conn(url, opts),
-           stream <- open_stream(pid, method, f_url, headers, body, true) do
+    defp request_stream(method, url, headers, body, opts),
+      do: do_request(method, url, headers, body, opts)
+
+    defp do_request(method, url, headers, body, opts) do
+      uri = URI.parse(url)
+      headers = headers ++ [{"host", uri.host}]
```
rinpatch: This will give wrong results when using non-standart port, no? I think instead you should just add a warning that gun <=1.3 (it's fixed in master) will append the port to the host header even if it's standart

lib/tesla/adapter/gun.ex
```diff
@@ -148,6 +148,20 @@ if Code.ensure_loaded?(:gun) do
...
+
+          tls_opts =
+            Map.get(opts, :tls_opts, %{})
+            |> Map.put(:server_name_indication, host)
+
+          Map.put(opts, :tls_opts, tls_opts)
+        else
+          opts
+        end
+
```
teamon: Any chances for test for this case?
Also, what does SNI has to do with FollowRedirects middleware? 🤔

alex-strizhakov: I think nope, because server_name_indication option is transferred to :ssl.connect and is not returned to the %Env{} in response. Gun opens tls connection through ip, that's why default sni option with host is not added by ssl module (http://erlang.org/doc/man/ssl.html#type-sni). SNI option is needed for some certificates, on which gun returns Handshake error. This option is tested on dev server :)

teamon:
👋 @alex-strizhakov  how's going? :)

alex-strizhakov:
hi @teamon !
Main options are here. What i have to do to make it merged? 
I can add tests for proxy option, by adding `tinyproxy` image, if it is needed. [How it is done in mint](https://github.com/ericmj/mint/tree/master/test/docker/tinyproxy)
SNI option can be removed after this [issue](https://github.com/ninenines/gun/issues/207) will be fixed.
Options for support `gun` 1.3 and master can be removed after release `gun` 2.

teamon:
Please take a look at the pending review comments :)

lib/tesla/adapter/gun.ex
alex-strizhakov: In new version it's made with [clause](https://github.com/teamon/tesla/pull/310/files#diff-f93babd4700a230f6ced434c9dd3a68eR148)

test/tesla/adapter/gun_test.exs
alex-strizhakov: After adding some tests it's used 5 times

test/tesla/adapter/gun_test.exs
alex-strizhakov: [Reused connection test with plain body](https://github.com/teamon/tesla/pull/310/files#diff-395315891a5107d753ddcfe9ab7998e5R83)
[Reading body in chunks with reusing connection](https://github.com/teamon/tesla/pull/310/files#diff-395315891a5107d753ddcfe9ab7998e5R109)
[Reading body as stream with reusing connection](https://github.com/teamon/tesla/pull/310/files#diff-395315891a5107d753ddcfe9ab7998e5R186)

lib/tesla/adapter/gun.ex
alex-strizhakov: Adding `host` header is removed.

lib/tesla/adapter/gun.ex
alex-strizhakov: Is it needed test for this [case](https://github.com/teamon/tesla/pull/310/files#diff-f93babd4700a230f6ced434c9dd3a68eR175)?

alex-strizhakov:
I thought they all were resolved :)

teamon:
Let's do it then! 🚀

alex-strizhakov:
👍

PR #313, opened 2019-08-23
Fix relative redirect support

myfreeweb:
Redirects can be relative to the current URL.

e.g.

```elixir
iex(sr2@test)3> Tesla.get("https://webmention.rocks/test/23/page")
{:ok,
 %Tesla.Env{
   # …
   headers: [
     {"location", "page/wjj46SaVj1OnFIzTBDwA"}, # …
   ],
   status: 302,
 }}
```

so it's the `^https?://` case where we can bypass URI parsing. In other cases, let `URI.merge` do its thing.

### DIFF
--- a/lib/tesla/middleware/follow_redirects.ex
+++ b/lib/tesla/middleware/follow_redirects.ex
@@ -68,12 +68,13 @@ defmodule Tesla.Middleware.FollowRedirects do
   defp new_request(%{status: 303} = env, location), do: %{env | url: location, method: :get}
   defp new_request(env, location), do: %{env | url: location}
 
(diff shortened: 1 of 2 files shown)


teamon:
I'm not sure this change actually changes implementation. The original one has

```ex
defp parse_location("/" <> _rest = location, env), do: ... parsing ...
defp parse_location(location, _env), do: location
```

If you call `parse_location("https://....")` it will not match the first clause, follow to the next one and return the passed location unchanged. 

What am I missing?

teamon:
Ok, nevermind, I got it 😅 

Thanks for the contribution! 👏

PR #320, opened 2019-09-25
Socks proxy support - gun adapter

alex-strizhakov:
- Removed adding SNI options
- added support for socks proxy

### DIFF
--- a/test/tesla/adapter/gun_test.exs
+++ b/test/tesla/adapter/gun_test.exs
@@ -237,6 +237,16 @@ defmodule Tesla.Adapter.GunTest do
     assert response.status == 500
   end
 
+  test "error on socks proxy" do
+    request = %Env{
+      method: :get,
+      url: "#{@http}/status/500"
+    }
+
+    assert {:error, "socks protocol is not supported"} ==
+             call(request, proxy: {:socks5, 'localhost', 1234})
+  end
+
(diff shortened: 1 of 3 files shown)


test/tesla/adapter/gun_test.exs
```diff
@@ -237,6 +237,16 @@ defmodule Tesla.Adapter.GunTest do
...
+  test "error on socks proxy" do
+    request = %Env{
+      method: :get,
+      url: "#{@http}/status/500"
+    }
+
+    assert {:error, "socks protocol is not supported"} ==
+             call(request, proxy: {:socks5, 'localhost', 1234})
+  end
+
```
teamon: can we get happy-path test somehow?

alex-strizhakov: I think no, because socks support is added to the gun in the master branch (2.0.0-pre.1 version). Tesla uses 1.3. Gun version can be bumped when gun 2.0 release will be ready.
Main problem in testing http/socks proxies is that we need to add proxy servers like [tinyproxy](https://tinyproxy.github.io) to the repo. Are you ok with that?

teamon: Ideally, adapter code should be as simple as possible so adding another test dependency might not be a good idea unless we can replace httparrot which seems to be not very reliable recently, especially when testing ssl (although I haven't looked into the reasons for that).

PR #322, opened 2019-10-02
Update Hex version in README

mmcc:
Silly mistake on my part, but I blindly copied the `deps` example and spent a few minutes debugging why the Mint adapter wasn't working. Updating the README example for future blind copiers :)


teamon:
Thanks!

PR #324, opened 2019-10-08
Conform telemetry middleware to established patterns

bryannaegele:
Hello! I realize this will be a breaking change for the telemetry middleware but it should be worth it.

With the changes in this PR, this middleware will mirror conventions we're using elsewhere in the ecosystem, allow for more telemetry to be captured (including exceptions), and enables Tesla clients to potentially be traced using telemetry events (TBD - we're exploring this)!

#### Background
The patterns we've established with telemetry modules which involve durations have coalesced with a few key characteristics:
* Events & Measurements:
  * emit a `:start` event with the current time as the measurement under the `:time` key.
  * emit a `:stop` event with the duration as the measurement under the `:duration` key.
* Metadata:
  * emit the main data structure under a key with that name, e.g. `%{conn: conn}` in `Plug.Telemetry` and `Phoenix`
* allow the event to have a custom prefix when it could be implemented multiple times so that each implementation may be differentiated
* treat the call to telemetry as a side effect by executing it in an `after` clause so as to not skew the results


teamon:
Any breaking change can only go to tesla 2.0.

Can you share a bit more on who is "we" and provide some references for "established patterns"?

I haven't looked into details yet, but for startes I would be rather against promoting per-client prefix - it makes it harder to provide a global metric capturing all outgoing comms from one app (but again, I haven't looked much into the details).

bryannaegele:
Hey @teamon! By "we" I was referring to various members of the [ErlEF Observability Working Group](https://erlef.org/wg/observability) and for "established patterns" I was referring to the patterns done from that work in:
* [Plug.Telemetry](https://hexdocs.pm/plug/Plug.Telemetry.html#content)
* [Phoenix](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#module-instrumentation)

Those show the `:start` and `:stop` conventions, as well as putting the `conn` object in the metadata as `%{conn: conn}` and using `:duration` for the measurement name. The `:duration` piece isn't a dealbreaker, but we're trying to consolidate on terminology. I'll bring it up that we need to put together a section for `telemetry` in the Library Guidelines. 👍 

For prefixing options, we have that in `Plug.Telemetry` and in [Ecto](https://hexdocs.pm/ecto/Ecto.Repo.html#module-telemetry-events). This would be optional here, so the user would be making a conscious choice. I'm open to other ideas as to how a consumer of the event could determine which client was called to get that into metadata, but that should be possible for purposes of tracing.

Some discussion on `:start` and `:stop` events, there's a lot of discussion from the team in [this issue](https://github.com/elixir-plug/plug/pull/824). The general idea is without a `:start` event, authors of tracing libraries would have to figure out some way to retroactively create child spans, which would be very complicated.

If these changes would need to wait for a major version release, I can create something internally for our use case while we work to a good solution. :)

lib/tesla/middleware/telemetry.ex
```diff
@@ -3,30 +3,98 @@ if Code.ensure_loaded?(:telemetry) do
...
+
+    defp emit_start(req_env, start_time, prefix) do
+      :telemetry.execute(prefix ++ [:tesla, :request, :start], %{time: start_time}, %{
+        env: req_env
+      })
+    end
+
+    defp emit_result(result, start_time, prefix, req_env) do
+      try do
+        result
```
teamon: What is this supposed to do? 🤔

teamon:
Thank your for this in-depth explanation, it is much clearer now  👏 

In general, I agree. It makes sense to send separate `:start` and `:stop` events.
At the same time, I take stability and backwards compatibility as the highest priority for this library, so we can't just break it.

One solution I can think of, is to send both: the new start/stop events and the old `[:tesla, :request]` one at the same time, with an option to disable the old one with opt-in compile-time configuration.

bryannaegele:
@teamon what do you think about continuing to emit the current telemetry event for backwards compatibility? All new adopters would attach handlers on the new events but the existing implementations wouldn't break. Whenever you do a new major version, you could just delete the old event.

lib/tesla/middleware/telemetry.ex
bryannaegele: I was avoiding nesting two of these in the main function. :)

Since `after` is evaluated after the try or catch bodies are evaluated, recording the end event after those bodies have run gives us the actual time that the operation took. Since it is possible for Tesla to raise an exception, the first `try do` in the main function would emit the event with the `env` being what was passed to the function (we can't read evaluated values from the `try do` block because it may not exist in the case of an exception being raised. 

In the `emit_result` function I was using the same pattern to emit the event as a side effect. This may not be absolutely necessary since we're just returning the value passed, however it wouldn't cause any harm, so I _think_ it would be ok to omit it if you have strong feelings about it.

teamon: Elixir is a strict language - arguments are evaluated before they are passed to functions. This `try` will not have any effect.

teamon:
Compile time could be something along these lines:


```elixir

if Application.get_env(:tesla, Tesla.Middleware.Telemetry, :smth) do
  defp emit_result(...) do 
    # emit the new way
  end
else
  defp emit_result(...) do
    # emit the legacy way
  end
end
```

I'd just like to avoid calling `Application.get_env` repeatedly (which would be the case when using runtime checks)

lib/tesla/middleware/telemetry.ex
bryannaegele: Correct. Normally we do everything in the `after` as a little trick but there's no guarantee the result would have been evaluated since `after` runs on success or on exception. Either way, this is a moot point with the changes I made last night to simplify it. 👍

bryannaegele:
I added an option to disable the legacy event at compile time. We always want to emit the new events, so I opted to just inline it into the existing function rather than writing two versions.

Failing to disable the event isn't a big deal. We're talking ~4µs of added time if users don't explicitly disable.

lib/tesla/middleware/telemetry.ex
```diff
@@ -3,30 +3,111 @@ if Code.ensure_loaded?(:telemetry) do
...
-    Send the request time and meta-information through telemetry.
+    Emits events using the `:telemetry` library to expose instrumentation.
 
     ### Example usage
     ```
     defmodule MyClient do
       use Tesla
 
-      plug Tesla.Middleware.Telemetry
+      plug Tesla.Middleware.Telemetry, prefix: [:my_client]
```
teamon: I'd like to make it 100% clear that this is optional - let's move this to the list of options and leave the default example without `prefix`

lib/tesla/middleware/telemetry.ex
```diff
@@ -3,30 +3,111 @@ if Code.ensure_loaded?(:telemetry) do
...
+      * Metadata: `%{env: Tesla.Env.t}`
+
+    * `[:tesla, :request, :error]` - emitted at the end of the request when there is an error.
+      * Measurement: `%{duration: native_time}`
+      * Metadata: `%{env: Tesla.Env.t, reason: term}`
+
+    * `[:tesla, :request, :exception]` - emitted at the end of the request when an exception is raised.
+      * Measurement: `%{duration: native_time}`
+      * Metadata: `%{env: Tesla.Env.t, exception: Exception.t, stacktrace: Exception.stacktrace}`
+      * `[:tesla, :request]` - This event is emitted for backwards compatibility only and should be considered deprecated.
```
teamon: Let's move this into separate `## Legacy Telemetry Events` section

lib/tesla/middleware/telemetry.ex
```diff
@@ -3,30 +3,111 @@ if Code.ensure_loaded?(:telemetry) do
...
+      This event can be disabled by setting `config :tesla, Tesla.Middleware.Telemetry, disable_legacy_event: true` in your config.
+
     Please check the [telemetry](https://hexdocs.pm/telemetry/) for the further usage.
     """
 
+    @disable_legacy_event Application.get_env(:tesla, Tesla.Middleware.Telemetry,
```
teamon: If we use module prefix we need to add a note to recompile tesla (`mix deps.compile --force tesla` ) when changing this config, otherwise it won't be picked up automatically.

lib/tesla/middleware/telemetry.ex
```diff
@@ -3,30 +3,111 @@ if Code.ensure_loaded?(:telemetry) do
...
+            :telemetry.execute(
+              prefix ++ [:tesla, :request, :exception],
+              %{duration: System.monotonic_time() - start_time},
+              metadata
+            )
+
+            reraise e, stacktrace
+        end
+
+      emit_stop(result, start_time, prefix, env)
```
teamon: Nitpick - I'd use `emit_stop(env, start_time, prefix, result)` to match the signature of `emit_start`

lib/tesla/middleware/telemetry.ex
```diff
@@ -3,30 +3,111 @@ if Code.ensure_loaded?(:telemetry) do
...
+    end
+
+    defp emit_start(env, start_time, prefix) do
+      :telemetry.execute(prefix ++ [:tesla, :request, :start], %{time: start_time}, %{
+        env: env
+      })
+    end
+
+    defp emit_stop(result, start_time, prefix, req_env) do
+      duration = System.monotonic_time() - start_time
```
teamon: Why call `System.monotonic_time()` inside `emit_stop` but outside of `emit_start`? It would be easier to reason if this was consistent (we can pass `duration` to `emit_stop` as it doesn't need `start_time` value)

lib/tesla/middleware/telemetry.ex
bryannaegele: Good call.

lib/tesla/middleware/telemetry.ex
bryannaegele: Fair. I added a note on the option to help people understand when it's useful to set.

bryannaegele:
Updated per feedback. I've asked some of the other wg folks to provide feedback before wrapping this up.

lib/tesla/middleware/telemetry.ex
```diff
@@ -14,19 +14,103 @@ if Code.ensure_loaded?(:telemetry) do
 
     end
 
-    :telemetry.attach("my-tesla-telemetry", [:tesla, :request], fn event, time, meta, config ->
+    :telemetry.attach("my-tesla-telemetry", [:tesla, :request, stop], fn event, measurements, meta, config ->
```
binaryseed: `:stop` ?

teamon:
@bryannaegele Looks like there is one tiny fix + tests and the PR will be ready to go.

bryannaegele:
Apologies @teamon. Events at work prevented me from wrapping this up.

I'll fix up the test shortly, but I'd like to shore up a few discussions with the working group on conventions for the exception case as it pertains to tracing before merging.

bryannaegele:
Updated per our latest determinations as to event naming, metatdata, and event executions in beam-telemetry/telemetry_metrics/issues/46

@GregMefford @josevalim @binaryseed please review. This implementation has some additional complexities with the legacy event, but the rest of the pattern we discussed should largely match.

lib/tesla/middleware/telemetry.ex
```diff
@@ -13,21 +13,102 @@ if Code.ensure_loaded?(:telemetry) do
...
+      * Measurement: `%{time: System.monotonic_time}`
+      * Metadata: `%{env: Tesla.Env.t}`
+
+    * `[:tesla, :request, :stop]` - emitted at the end of the request.
+      * Measurement: `%{duration: native_time}`
+      * Metadata: `%{env: Tesla.Env.t}`
+
+    * `[:tesla, :request, :error]` - emitted when there is an error.
+      * Measurement: `%{value: 1}`
+      * Metadata: `%{env: Tesla.Env.t, kind: Exception.kind | nil, reason: term, stacktrace: Exception.stacktrace}`
```
josevalim: Hrm, should we make a distinction between error tuples and exceptions? It feels weird to put them in the same bucket but I am not sure which approach is effectively better. Which kind of errors do we have in the error tuple? Connection errors? What about a 500 response? Is that still an OK tuple?

lib/tesla/middleware/telemetry.ex
```diff
@@ -13,21 +13,102 @@ if Code.ensure_loaded?(:telemetry) do
...
+          stacktrace = System.stacktrace()
+          metadata = %{env: env, kind: kind, reason: reason, stacktrace: stacktrace}
+
+          :telemetry.execute(
+            prefix ++ [:tesla, :request, :error],
+            %{value: 1},
+            metadata
+          )
+
+          emit_stop(env, start_time, prefix, {:error, reason})
```
josevalim: In this case we emit `:error` and no `:stop` but in the error tuple below, we emit both `:error` and `:stop`. I think we the consensus was to always emit `:stop`.

lib/tesla/middleware/telemetry.ex
bryannaegele: I had an exception event in the [initial design](teamon/tesla/pull/324/commits/c1e01c1a13a9ab9a982cdfd00c69cce16b6c14c9) to distinguish but prior art in tracing seems to favor lumping them together. My initial inclination was the same as yours.

https://opentracing.io/docs/overview/tags-logs-baggage/
https://opentracing.io/specification/conventions/

@tsloughter may have a better handle on if these are being carried into OTel (I'm guessing so for backward compatibility).

>Which kind of errors do we have in the error tuple? Connection errors? What about a 500 response? Is that still an OK tuple?

That is a great question. API designers are going to handle these differently. @teamon can correct me if I'm wrong but, in Tesla's case, an ok tuple is returned for any request which completes and all error tuples are the result of a failure to complete the request, including timeouts, adapter failures, DNS lookup failures, etc.

One issue I'm seeing with establishing a good pattern is passing the result to stop as metadata. If we try to do this in the after, we don't have a way to access that result. This is probably OK in some cases, but I've found that the metadata is frequently used in our metrics in the wild. I'm struggling to figure out a pattern that doesn't require dropping the after clause and writing out more granular events for each possible situation.

lib/tesla/middleware/telemetry.ex
bryannaegele: The error is emitted on line 67. The `emit_stop/4` with that last argument is only there for backward compatibility with the previous implementation, which emitted the result tuple as metadata. I was creating an error tuple to pass to it to fit the previous pattern.

I wonder if it may be more useful to just be more explicitly verbose and mark the deprecated pieces to be deleted later in comments rather than this function.

binaryseed:
I know the prefix pattern exists elsewhere but it does make it harder to auto-discover & instrument w/o extra configuration done by the user (which is something vendors like New Relic try to do)...

Is there something we could do that would enable discovering the prefix via code? Ecto sets them via `config` which makes it possible to read in...

josevalim:
@binaryseed there is an open issue on Telemetry to provide some sort of discovery. It can also be done by tracing `telemetry.execute` calls.

Anyway, I support removing the prefixes for now and add them if necessary. I would expect most libs using Tesla to provide their own telemetry in addition to Tesla's. For example, a Twitter client could provide its own events and metadata.

bryannaegele:
I have a potential solution to the prefix question which we can discuss in this weeks meeting.

As to the prefix, I'm open to suggestions but I do not see a way in which a user can distinguish between clients without completely reimplementing this middleware on their own. I want to make sure we aren't trying to only solve for tracing where we can add context to a wrapping span.

The requirement I'm trying to make sure this solution would solve is we have code today where we have multiple Tesla clients which have different latency tolerances for the action they are performing. For some, approaching 5 seconds is cause for concern while for another it's 20ms. Our bucketing for Prometheus are tight around those times to more accurately capture those tolerances to our p99s. 

Our only way for distinguishing between telemetry events is via the event name. Telemetry handlers are tied only to event names and Telemetry.Metrics provides no mechanism for filtering measurements based upon values of metadata properties. Users should be able to have a mechanism by which they can distinguish between telemetry events from various implementations of a piece of code, be it a library or otherwise.

I am open to alternative suggestions and could be convinced otherwise, but I don't believe users should be forced to reimplement this middleware to correlate instances of the library to their measurements.

lib/tesla/middleware/telemetry.ex
```diff
@@ -13,21 +13,102 @@ if Code.ensure_loaded?(:telemetry) do
...
 
-    :telemetry.attach("my-tesla-telemetry", [:tesla, :request], fn event, time, meta, config ->
+    :telemetry.attach("my-tesla-telemetry", [:tesla, :request, :stop], fn event, measurements, meta, config ->
       # Do something with the event
     end)
     ```
 
+    ## Options
+
+    * `:event_prefix` - a list of atoms to prefix to the telemetry event name. This can be set if you need to distinguish events from different clients. Defaults to `[]`
```
GregMefford: I wonder if there would be a way to put this into the `meta` instead (like `client: :my_client` field) so that the event name itself isn't dynamic, but you could still pattern-match on the metadata if you need to.

bryannaegele:
It took a long road but the finalized standard and conventions feel good.

@GregMefford @josevalim @arkgil @tsloughter @binaryseed please review

lib/tesla/middleware/telemetry.ex
```diff
@@ -13,21 +13,101 @@ if Code.ensure_loaded?(:telemetry) do
...
+
+    * `[:tesla, :request, :start]` - emitted at the beginning of the request.
+      * Measurement: `%{time: System.monotonic_time}`
+      * Metadata: `%{env: Tesla.Env.t()}`
+
+    * `[:tesla, :request, :stop]` - emitted at the end of the request.
+      * Measurement: `%{duration: native_time}`
+      * Metadata: `%{env: Tesla.Env.t()} | %{env: Tesla.Env.t, error: term}`
+
+    * `[:tesla, :request, :fail]` - emitted when there is an error.
```
josevalim: Sorry but I just proposed for us to name this `:failure` instead of `:fail`. I noticed `:fail` reads a bit weird when implementing the same on Broadway. Please let me know your thoughts. :)

lib/tesla/middleware/telemetry.ex
```diff
@@ -13,21 +13,101 @@ if Code.ensure_loaded?(:telemetry) do
...
+      * Measurement: `%{value: 1}`
```
binaryseed: What is this intended to be?

A duration on the failure would be nice too, since it could be a timeout, etc...

bryannaegele: That's a good callout since we're standardizing on this being one of the two "closing" events. We should call that out in the spec, also

binaryseed: What is `value` here?

lib/tesla/middleware/telemetry.ex
bryannaegele: With the metadata filtering, it will be possible to figure out which implementation it was using the hidden property `__module__` in the `env`.

lib/tesla/middleware/telemetry.ex
```diff
@@ -13,21 +13,102 @@ if Code.ensure_loaded?(:telemetry) do
...
-    :telemetry.attach("my-tesla-telemetry", [:tesla, :request], fn event, time, meta, config ->
+    :telemetry.attach("my-tesla-telemetry", [:tesla, :request, :stop], fn event, measurements, meta, config ->
       # Do something with the event
     end)
     ```
 
+    ## Telemetry Events
+
+    * `[:tesla, :request, :start]` - emitted at the beginning of the request.
+      * Measurement: `%{time: System.monotonic_time}`
```
binaryseed: Start time should **not** be a `monotonic_time`... Monotonic time is needed to accurately calculate durations so it needs to be captured for calculating `duration`, but it can not be converted into an actual timestamp.

Start time needs to be a `System.system_time` otherwise we won't be able to determine the actual beginning of the span

tsloughter: I was about to say the same thing :)

tsloughter: I wondered if they meant for system time to be taken in the handler and the `time` metadata to be used for duration?

josevalim: @tsloughter - correct. But then, theoretically speaking, we don't need to measure the duration either, because that could also be done in the handler. So where do we draw the line?

josevalim: Maybe we should call these fields "monotonic_time" or hide them altogether, as they are bound to be misused. Thoughts?

tsloughter: Definitely should at least name it `monotonic_time` and not just `time`.

josevalim: Moving the discussion to: https://github.com/beam-telemetry/telemetry/issues/57#issuecomment-587668574

binaryseed: In the absinthe instrumentation I did, I never put a monotonic time inside the emitted events, just the calculated `duration` and the system time for `start`, all in native time units.

I think it's reasonable to keep the monotonic time handling inside the instrumentation since it's not really obvious how to do it right.

binaryseed: I also generated a unique ID for each span and put that on all the events so that it's possible to correlate `start` with it's `stop` / `failure` w/o making the handler do a bunch of manual tracking

bryannaegele: Why do we need wall clock times this deep in a trace? This is the same behavior we use everywhere. 

https://hexdocs.pm/plug/Plug.Telemetry.html#module-time-unit

binaryseed: Yeah, Plug is "incorrect" if anyone is trying to turn that into an actual wall clock start time. 

Monotonic time tracks duration:
```
iex()> start_time = System.monotonic_time
-576460589343512497
iex()> end_time = System.monotonic_time
-576460587151522497
iex()> System.convert_time_unit(end_time - start_time, :native, :millisecond)
2191
```

But has nothing to do with wall clock time:
```
iex()> System.convert_time_unit(System.monotonic_time, :native, :millisecond)
-576460434182
```

The actual start time of a span is critical no matter where it is in the trace, I don't know how you'd visualize it or put the assemble trace back together without it

bryannaegele:
@josevalim @GregMefford - last updates in. Any objection to getting this shipped?

josevalim:
Looks good to me!
-- 


*José Valimhttps://dashbit.co/ <https://dashbit.co/>*

PR #326, opened 2019-10-12
add error handling to hackney adapter

### LINKED ISSUE #314: Hackney adapter does not handle {:error, :closed} returned by :hackeny.send_body/2
I am getting random `{:error, :closed}` from `:hackeny.send_body/2` and Tesla.client crashes with following stack trace.

```
 {:EXIT, #PID<0.23655.28>, {{:badmatch, {:error, :closed}}, [{
Tesla.Adapter.Hackney, :"-request_stream/5-fun-0-", 3, [file: 'lib/tesla/adapter/hackney.ex', line: 74]}, 
{Enum, :"-reduce/3-anonymous-2-", 3, [file: 'lib/enum.ex', line: 3015]}, 
{Stream, :do_transform_each, 3, [file: 'lib/stream.ex', line: 966]}, 
{Stream, :do_resource, 5, [file: 'lib/stream.ex', line: 1394]}, 
{Stream, :do_enum_transform, 7, [file: 'lib/stream.ex', line: 936]}, 
{Enum, :reduce, 3, [file: 'lib/enum.ex', line: 3015]}, 
{Tesla.Adapter.Hackney, :request_stream, 5, [file: 'lib/tesla/adapter/hackney.ex', line: 74]}, 
{Tesla.Adapter.Hackney, :call, 2, [file: 'lib/tesla/adapter/hackney.ex', line: 31]}
]}}
```

```
** (MatchError) no match of right hand side value: {:error, :closed}
(tesla) lib/tesla/adapter/hackney.ex:74: anonymous fn/3 in Tesla.Adapter.Hackney.request_stream/5
(elixir) lib/enum.ex:3015: anonymous fn/3 in Enum.reduce/3
(tesla) lib/tesla/adapter/hackney.ex:31: Tesla.Adapter.Hackney.call/2
(elixir) lib/stream.ex:1394: Stream.do_resource/5
(tesla) lib/tesla/adapter/hackney.ex:74: anonymous fn/3 in Tesla.Adapter.Hackney.request_stream/5
(elixir) lib/stream.ex:936: Stream.do_enum_transform/7
```

I am trying to make a `POST` request using `Tesla.post` using Tesla.Multipart body. 
**Method:** :post
**URL:** "server url"
**Headers:** [
  {"authorization", "Bearer <token>"},
  {"content-type", "multipart/form-data; boundary=<end_string>; charset=utf-8"}
]
**Opts:** [recv_timeout: 1800000]
**Multipart body:** 
```
Multipart.new
    |> Multipart.add_content_type_param("charset=utf-8")
    |> Multipart.add_field("field", <encoded_field_data>)
    |> Multipart.add_file(<path>,
         name: "filename",
         filename: Path.basename(<path>),
         headers: [{"content-type", "application/octet-stream"}]
       )
```

Dreat:
Now `:hackney.send_body/2` is wrapped with error handling
and will not crash with {:error, :closed} return (or any other than :ok)

This closes #314

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -72,21 +72,23 @@ if Code.ensure_loaded?(:hackney) do
 
     defp request_stream(method, url, headers, body, opts) do
       with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
-        res =
-          Enum.reduce_while(body, :ok, fn data, _ ->
-            case :hackney.send_body(ref, data) do
-              :ok -> {:cont, :ok}
-              error -> {:halt, error}
-            end
-          end)
-
-        case res do
+        case send_body(body ,ref) do
           :ok -> handle(:hackney.start_response(ref))
           error -> handle(error)
         end
       else
         e -> handle(e)
       end
+    end
+
+    defp send_body(body, ref) do
+      body
+      |> Enum.reduce_while(:ok, fn data, _ ->
+        case :hackney.send_body(ref, data) do
+          :ok -> {:cont, :ok}
+          error -> {:halt, error}
+        end
+      end)
     end
 
     defp handle({:error, _} = error), do: error
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -72,7 +72,14 @@ if Code.ensure_loaded?(:hackney) do
 
     defp request_stream(method, url, headers, body, opts) do
       with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
-        for data <- body, do: :ok = :hackney.send_body(ref, data)
+        for data <- body do
+          with :ok <- :hackney.send_body(ref, data) do
+            :ok
+          else
+            e -> handle(e)
+          end
+        end
+
         handle(:hackney.start_response(ref))
       else
         e -> handle(e)
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -72,15 +72,18 @@ if Code.ensure_loaded?(:hackney) do
 
     defp request_stream(method, url, headers, body, opts) do
       with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
-        for data <- body do
-          with :ok <- :hackney.send_body(ref, data) do
-            :ok
-          else
-            e -> handle(e)
-          end
-        end
+        res =
+          Enum.reduce_while(body, :ok, fn data, _ ->
+            case :hackney.send_body(ref, data) do
+              :ok -> {:cont, :ok}
+              error -> {:halt, error}
+            end
+          end)
 
-        handle(:hackney.start_response(ref))
+        case res do
+          :ok -> handle(:hackney.start_response(ref))
+          error -> handle(error)
+        end
       else
         e -> handle(e)
       end


lib/tesla/adapter/hackney.ex
```diff
@@ -72,7 +72,14 @@ if Code.ensure_loaded?(:hackney) do
...
       with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
-        for data <- body, do: :ok = :hackney.send_body(ref, data)
+        for data <- body do
+          with :ok <- :hackney.send_body(ref, data) do
+            :ok
+          else
+            e -> handle(e)
+          end
+        end
+
```
teamon: This won't work as you expect, given the `for ...` above - it will still try to send every body piece even after an error occurs, and discard the return value(s) of `handle(e)` and still call `handle(:hackney.start_response(ref))`

teamon:
Hey @Dreat , how's going? :)

Dreat:
@teamon life happened and I got distracted with different things

I though about how to do it, and will submit update on Monday/Tuesday. Sadly the only way I see it work is to convert Stream to Enum when sending body. But it will work :)

teamon:
No worries :)

I think it should be possible to implement without converting stream to enum using recursive function to send the first stream element if present (and than the next and next etc) or finish sending when empty - this way it will be possible to check for error on each send operation.

lib/tesla/adapter/hackney.ex
```diff
@@ -72,8 +72,18 @@ if Code.ensure_loaded?(:hackney) do
...
+            case :hackney.send_body(ref, data) do
+              :ok -> {:cont, :ok}
+              error -> {:halt, error}
+            end
+          end)
+
+        case res do
+          :ok -> handle(:hackney.start_response(ref))
+          error -> handle(error)
+        end
```
teamon: We could extract sending into separate function and then use it like:

```elixir
with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
    :ok <- send_body(ref, body)
  handle(:hackney.start_response(ref))
else
  e -> handle(e)
end
```

teamon:
Nice!

Can you think of a test we could write for this? Maybe we can make `:hackney.send_body` fail with some invalid stream element?

lib/tesla/adapter/hackney.ex
Dreat: Will extract it to function then :)
But I think `with` won't work here as we return accumulator (this time only :ok) or error.

Dreat:
@teamon I was thinking about test, but to be honest Hackney will fail when there's real sending error. I looked into source code and didn't find a way to make if fail on purpose.
I'm not that experienced with sending invalid streams thou, so I will be very glad on any headsup :)

teamon:
@Dreat I peeked inside hackney code and we can use this to fake `send_body` error:

```elixir
  test "stream request body: error" do
    body =
      Stream.unfold(5, fn
        0 -> nil
        3 -> {fn -> {:error, :fake_error} end, 2}
        n -> {to_string(n), n - 1}
      end)

    request = %Env{
      method: :post,
      url: "#{@http}/post",
      body: body
    }

    assert {:error, :fake_error} = call(request)
  end
```

Dreat:
> @Dreat I peeked inside hackney code and we can use this to fake `send_body` error:
> 
> ```elixir
>   test "stream request body: error" do
>     body =
>       Stream.unfold(5, fn
>         0 -> nil
>         3 -> {fn -> {:error, :fake_error} end, 2}
>         n -> {to_string(n), n - 1}
>       end)
> 
>     request = %Env{
>       method: :post,
>       url: "#{@http}/post",
>       body: body
>     }
> 
>     assert {:error, :fake_error} = call(request)
>   end
> ```

This test didn't work, `:hackney` will still return `:ok` here :(

teamon:
Hmmm I've tried that and it seemed to work 🤔 

@benoitc could you give us a hand here? We are trying to write a test case when body argument passed to `hackney:send_body/2` returns an error, so we can make one of the stream elements cause an error and ensure that error will bubble up.

xinz:
Hello, I did some check in`hackney` and the following test case works for me as well.

```elixir
test "stream request body: error" do
    body =
      Stream.unfold(5, fn
        0 -> nil
        3 -> {fn -> {:error, :fake_error} end, 2}
        n -> {to_string(n), n - 1}
      end)

    request = %Env{
      method: :post,
      url: "#{@http}/post",
      body: body
    }

    assert {:error, :fake_error} = call(request)
  end
```

@teamon I submitted another PR https://github.com/teamon/tesla/pull/368 base on this, could you please review and advise to resolve this issue?

benoitc:
i am not sure what you were trying to do but there is also function to stream the body in hackney.Maybe that could help?

xinz:
Hi @benoitc , thanks, I want to make sure your mentioned hackney's function to stream the (response) body is [stream_body/1](https://github.com/benoitc/hackney/blob/master/doc/hackney.md#stream_body1), right?

benoitc:
> Hi @benoitc , thanks, I want to make sure your mentioned hackney's function to stream the (response) body is [stream_body/1](https://github.com/benoitc/hackney/blob/master/doc/hackney.md#stream_body1), right?

yes :)

xinz:
Got it, thank you for your prompt reply ：）

teamon:
Replaced with https://github.com/teamon/tesla/pull/368

PR #333, opened 2019-11-12
Gun

alex-strizhakov:
Support for ipv4 & ipv6

### DIFF
--- a/lib/tesla/adapter/gun.ex
+++ b/lib/tesla/adapter/gun.ex
@@ -154,13 +154,7 @@ if Code.ensure_loaded?(:gun) do
     end
 
     defp check_original(%URI{host: host, port: port}, %{original: original} = opts) do
-      formatted_host =
-        case format_host(host) do
-          {:domain, domain} -> domain
-          {:ip, _ip} -> host
-        end
-
-      Map.put(opts, :original_matches, original == "#{formatted_host}:#{port}")
+      Map.put(opts, :original_matches, original == "#{domain_or_fallback(host)}:#{port}")
     end
 
     defp check_original(_uri, opts), do: opts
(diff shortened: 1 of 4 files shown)


lib/tesla/adapter/gun.ex
```diff
@@ -200,16 +206,19 @@ if Code.ensure_loaded?(:gun) do
 
       tls_opts =
         with "https" <- uri.scheme,
-             true <- opts[:certificates_verification],
-             host <- to_charlist(uri.host) do
-          host = :idna.encode(host)
+             true <- opts[:certificates_verification] do
+          formatted_host =
+            case format_host(uri.host) do
+              {:domain, domain} -> domain
```
teamon: `domain` is charlist or string?

lib/tesla/adapter/gun.ex
```diff
@@ -420,5 +429,17 @@ if Code.ensure_loaded?(:gun) do
...
+      charlist = to_charlist(string)
+
+      case :inet.parse_address(charlist) do
+        {:error, :einval} ->
+          {:domain, :idna.encode(charlist)}
+
+        {:ok, ip} when is_tuple(ip) and tuple_size(ip) in [4, 8] ->
+          {:ip, ip}
+      end
+    end
```
teamon: Would it make sense to have `domain_or_ss_a_string(uri)` instead? (name TBD) 

Seems like all use cases are to get encoded domain OR fallback to `uri.host` 🤔

test/support/adapter_case.ex
```diff
@@ -2,8 +2,14 @@ defmodule Tesla.AdapterCase do
...
-      @http "http://0.0.0.0:#{Application.get_env(:httparrot, :http_port)}"
-      @https "https://0.0.0.0:#{Application.get_env(:httparrot, :https_port)}"
+      # Needed for SSL test with certificate verification
+      if @adapter == Tesla.Adapter.Ibrowse do
+        @http "http://0.0.0.0:#{Application.get_env(:httparrot, :http_port)}"
+        @https "https://0.0.0.0:#{Application.get_env(:httparrot, :https_port)}"
+      else
+        @http "http://localhost:#{Application.get_env(:httparrot, :http_port)}"
+        @https "https://localhost:#{Application.get_env(:httparrot, :https_port)}"
+      end
```
teamon: It would be better to override this in Ibrowse tests module

lib/tesla/adapter/gun.ex
alex-strizhakov: `domain` here must be charlist

lib/tesla/adapter/gun.ex
alex-strizhakov: I renamed `format_host` function to return domain or ip. 
Created second function to return domain or fallback for ips. As `gun` wants domain or ip tuple to connect, but in other places needed fallback for ips.

test/support/adapter_case.ex
alex-strizhakov: Done 🙂

PR #341, opened 2019-12-11
When requests fails (while using timeout addition), preserve original stacktrace

esse:
Otherwise the stacktrace points to;

```
(tesla) lib/tesla/middleware/timeout.ex:54: Tesla.Middleware.Timeout.repass_error/1
(tesla) lib/tesla/middleware/timeout.ex:32: Tesla.Middleware.Timeout.call/3
(...)
```

which isn't very helpful.

### DIFF
--- a/lib/tesla/middleware/timeout.ex
+++ b/lib/tesla/middleware/timeout.ex
@@ -45,7 +45,7 @@ defmodule Tesla.Middleware.Timeout do
         {:ok, func.()}
       rescue
         e in _ ->
-          {:exception, e, __STACKTRACE__}
+          {:exception, e, System.stacktrace()}
       catch
         type, value ->
           {type, value}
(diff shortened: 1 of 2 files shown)


teamon:
👋  Thank you for your contribution!

I'm all for improving the stacktrace, as long as we do **not** hide an important trace.
We would definitely need tests for this.

Btw, `__STACKTRACE__` is available only in elixir 1.7+ while tesla supports `"~> 1.5"`.

esse:
@teamon thanks for your feedback!

I've changed `__STACKTRACE__` to `System.stacktrace/0` for compatibility and I've added tests for expected behaviour.

esse:
@teamon can you look into this? It would be very helpful for us to have this merged, since for now we're using forked version.

teamon:
Please rebase on master and ensure tests are passing :)

esse:
@teamon - thanks, done 👍

PR #342, opened 2019-12-11
Add option of having baseurl as a lambda

esse:
Which will make crafting releases easier.

### DIFF
--- a/lib/tesla/middleware/core.ex
+++ b/lib/tesla/middleware/core.ex
@@ -39,6 +39,8 @@ defmodule Tesla.Middleware.BaseUrl do
     end
   end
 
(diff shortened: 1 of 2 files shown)


teamon:
You can do this with [Runtime middleware](https://github.com/teamon/tesla#runtime-middleware)

```elixir
defmodule GitHub do
  # notice there is no `use Tesla`

  def issues(client \\ new()) do
    Tesla.get(client, "/issues")
  end

  # build dynamic client based on runtime arguments
  def new do
    middleware = [
      {Tesla.Middleware.BaseUrl, dynamic_base_url},
      Tesla.Middleware.JSON,
      ...
    ]

    Tesla.client(middleware)
  end
end
```

If we allow `base_url` as function we would have to add that option to every single option, which does not scale.

PR #353, opened 2020-02-06
Change regular expression in PathParms middleware to avoid replacing port in url

asniaire:
When `PathParams` middleware replaces the path params, it fails if the url contains the host with port (http://localhost: 7473/my_path/:whatever)

```elixir
** (ArgumentError) argument error
    :erlang.binary_to_existing_atom("7473", :utf8)
    (stdlib) erl_eval.erl:680: :erl_eval.do_apply/6
    (stdlib) erl_eval.erl:888: :erl_eval.expr_list/6
    (stdlib) erl_eval.erl:411: :erl_eval.expr/5
    (stdlib) erl_eval.erl:273: :erl_eval.expr/5
    (elixir) lib/regex.ex:731: Regex.apply_list/5
    (elixir) lib/regex.ex:726: Regex.apply_list/5
    (elixir) lib/regex.ex:668: Regex.do_replace/4
```

### Problem
It's unavoidable when using `BaseUrl`, `Telemetron.Instrumenter.Tesla` and `PathParams`. There is no order possible.

`Telemetron` needs `BaseUrl` to metric `host`, `PathParams` needs to be executed before `BaseUrl` because otherwise it would parse the port and `Telemetron` needs to be executed before `PathParams` because otherwise it would provoke high cardinality in metrics.

### Solution
Change regular expression to avoid considering numbers

### DIFF
--- a/lib/tesla/middleware/path_params.ex
+++ b/lib/tesla/middleware/path_params.ex
@@ -24,7 +24,7 @@ defmodule Tesla.Middleware.PathParams do
 
   @behaviour Tesla.Middleware
 
-  @rx ~r/:([\w_]+)/
+  @rx ~r/:([a-zA-Z_]+)/
 
   @impl Tesla.Middleware
   def call(env, next, _) do
--- a/lib/tesla/middleware/path_params.ex
+++ b/lib/tesla/middleware/path_params.ex
@@ -24,7 +24,7 @@ defmodule Tesla.Middleware.PathParams do
 
   @behaviour Tesla.Middleware
 
-  @rx ~r/:([a-zA-Z_]+)/
+  @rx ~r/:([a-zA-Z]{1}[\w_]*)/
 
   @impl Tesla.Middleware
   def call(env, next, _) do
--- a/test/tesla/middleware/path_params_test.exs
+++ b/test/tesla/middleware/path_params_test.exs
@@ -34,4 +34,28 @@ defmodule Tesla.Middleware.PathParamsTest do
 
     assert env.url == "/users/1/p/2"
   end
+
+  test "placeholder start by number" do
+    opts = [path_params: ["1id": 1, id_post: 2]]
+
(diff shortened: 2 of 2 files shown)


asniaire:
> What about `:my_param_123` ?

You're right. Would it be acceptable if we force to start by character?

Something like: `:([a-zA-Z]{1}[\w_]+)`

Otherwise we could opt by another approach:

```elixir
defp build_url(url, nil), do: url

  defp build_url(url, params) do
    Enum.reduce(params, url, fn {k, v}, u ->
      String.replace(u, ":#{k}", to_string(v))
    end)
  end
```

teamon:
Please see https://github.com/teamon/tesla/pull/270#issuecomment-447933951 on why `String.replace` is not a good solution.

`:([a-zA-Z]{1}[\w_]+)` will [not match single letter param](https://rubular.com/r/GGcaRiM8LFxq0o).

asniaire:
> Please see [#270 (comment)](https://github.com/teamon/tesla/pull/270#issuecomment-447933951) on why `String.replace` is not a good solution.

I see, thanks for this! Fair enough.

> `:([a-zA-Z]{1}[\w_]+)` will [not match single letter param](https://rubular.com/r/GGcaRiM8LFxq0o).

Arg, right! I forgot this! So, what about `:([a-zA-Z]{1}[\w_]*)`?

teamon:
Thanks! How about some tests? :)

asniaire:
> Thanks! How about some tests? :)

Sure, sorry! 🤦‍♂

Now it's done! Let me know if you think it covers all the cases we've discussed :)

> Note: After adding a few additional tests the pipeline failed because of a test not related (apparently) to my tests.

asniaire:
Well, after adding a last commit it passed the checks again 🤷‍♂

asniaire:
Thanks! 🙇

Issue #351, opened 2020-02-06
Make Fuse return original error when circuit is closed


asniaire:
## Problem
`Fuse` middleware is returning `{:error, :unavailable}` when circuit is open (expected) 
https://github.com/teamon/tesla/blob/master/lib/tesla/middleware/fuse.ex#L58

But also when the call has failed, hidden the original error.
https://github.com/teamon/tesla/blob/master/lib/tesla/middleware/fuse.ex#L73

I think it would be interesting to return the original error or maybe include a third element in the tuple including the original element.

For me, it's interesting to know when the circuit is open and the request never takes place, and know when the call has been performed with error as result.

## Timeout middleware integration
I ask for this after using `Timeout` and `Fuse` middlewares. For me, it makes sense to keep the order `Fuse` -> `Timeout`, so that if there is a timeout error, Fuse is aware of this when calculating its circuit status. But doing it, the original `{:error, :timeout}` is hidden by Fuse, returning `{:error, :unavailable}`. And I'd like to know that `timeout` was the original problem.

asniaire:
Another point I've realised is that, if `Tesla` gets a `5xx` error it returns `{:ok, %Tesla.Env{status: 5xx}` and it seems it's taken by `Fuse` as a successful response. So, it's not considered to melt the fuse. That is, to open the circuit.

Is it right? If so, is it the expected behaviour for a `circuit breaker`?

And maybe even http status `429` (`Too many requests`) should be considered by `Fuse`?

Thanks!

teamon:
It's impossible to have one-size-fits-all answer for "to melt or not to melt".
For the same reasons Retry middleware exposes (`should_retry` function](https://github.com/teamon/tesla/blob/master/lib/tesla/middleware/retry.ex#L32), Fuse middleware could use the same.

asniaire:
Yes! It was just was I was about to propose. Having an optional configuration to let the client decide which http status you want to consider to melt.

If you agree, I'll submit an MR containing this proposal.

Not sure if do exactly the same as `Retry` of just passing a list of http status codes 🤔

teamon:
Let's do the same as in Retry for consistency (also list of http status codes might not be enough)

asniaire:
Sure! It was just to make the clients simpler. But let's do the same way as `Retry` for consistency, yes :)

asniaire:
[...]

PR #354, opened 2020-02-07
Improve Fuse with custom melt conditions

asniaire:
Issue #351 

This `PR` also makes `Fuse` return the original response if the circuit is closed (fuse is not melted)

### DIFF
--- a/lib/tesla/middleware/fuse.ex
+++ b/lib/tesla/middleware/fuse.ex
@@ -31,7 +31,7 @@ if Code.ensure_loaded?(:fuse) do
 
     - `:name` - fuse name (defaults to module name)
     - `:opts` - fuse options (see fuse docs for reference)
-    - `:should_melt` - function to determine if request should melt the fuse
+    - `:should_melt` - function to determine if response should melt the fuse
 
     ## SASL logger
 
--- a/lib/tesla/middleware/fuse.ex
+++ b/lib/tesla/middleware/fuse.ex
@@ -19,7 +19,7 @@ if Code.ensure_loaded?(:fuse) do
 
       plug Tesla.Middleware.Fuse,
         opts: {{:standard, 2, 10_000}, {:reset, 60_000}},
-        keep_orig_error: true,
+        keep_original_error: true,
         should_melt: fn
           {:ok, %{status: status}} when status in [428, 500, 504] -> true
           {:ok, _} -> false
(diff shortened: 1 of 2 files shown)


asniaire:
It seems we do have a backwards compatibility issue in tests `v1.0.0` :(

I guess this is running current tests against `v1.0.0`, isn't it?

What am I supposed to do here?

lib/tesla/middleware/fuse.ex
```diff
@@ -48,30 +55,36 @@ if Code.ensure_loaded?(:fuse) do
...
 
-        {:error, _reason} ->
-          :fuse.melt(name)
-          {:error, :unavailable}
+    defp maybe_melt_fuse(res, context) do
+      if context.should_melt.(res) do
+        :fuse.melt(context.name)
       end
+
+      res
```
teamon: Why do we need a separate function here? How about 

```elixir
def run(env, next, %{should_melt: should_melt, name: name}) do
  res = Tesla.run(env, next)

  if should_melt.(res), do: :fuse.melt(name)

  res
end
```

?

asniaire: Sure, not really needed. Indeed, your suggestion was my initial implementation xD

Let me do it this way :)

teamon:
After sleeping on it, I think it would be better if we keep the backwards compatibility. 
We can however add a opt-in param to the middleware to return original error and document that this will be the default in tesla 2.0.0.

asniaire:
> After sleeping on it, I think it would be better if we keep the backwards compatibility.
> We can however add a opt-in param to the middleware to return original error and document that this will be the default in tesla 2.0.0.

I think this is the best solution too :) 

So, is it enough if this param is something like `keep_orig_resp: true | false` (default: `false`, but `true` in `2.0.0`)? Or what works best for you?

Thanks!

test/tesla/middleware/fuse_test.exs
```diff
@@ -67,7 +68,7 @@ defmodule Tesla.Middleware.FuseTest do
   end
 
   test "custom should_melt function - not melting 500" do
-    custom_client = ClientWithCustomShouldMeltFunction
+    custom_client = ClientWithCustomSetup
```
teamon: Let's get rid of this variable completely, it's quite confusing.

lib/tesla/middleware/fuse.ex
```diff
@@ -18,7 +18,8 @@ if Code.ensure_loaded?(:fuse) do
       use Tesla
 
       plug Tesla.Middleware.Fuse,
-        opts: {{:standard, 2, 10_000}, {:reset, 60_000}}
+        opts: {{:standard, 2, 10_000}, {:reset, 60_000}},
+        keep_orig_error: true,
```
teamon: `keep_original_error` maybe? 😉

asniaire: Yep! Better :)

test/tesla/middleware/fuse_test.exs
asniaire: Sure!

asniaire:
Thanks @teamon! 🙇

PR #359, opened 2020-02-19
Include `Tesla.Adapter.Mint` timeout's default value in documentation

tegon:
I thought this might be helpful for someone else in the future. I spent some time looking for the default value for `timeout` in Hexdocs until I found it in the source code.
I've also mentioned that the value is in `milliseconds` to be more explicit.

Let me know what you think. Thanks!


tegon:
Not sure why one of the builds failed since this change is only touching documentation.

PR #361, opened 2020-03-03
Strip Authorization Headers on Redirect

lizziepaquette:
Same issue as https://github.com/myfreeweb/httpotion/issues/135 

We shouldn't forward secrets!

### DIFF
--- a/lib/tesla/middleware/follow_redirects.ex
+++ b/lib/tesla/middleware/follow_redirects.ex
@@ -26,7 +26,9 @@ defmodule Tesla.Middleware.FollowRedirects do
   def call(env, next, opts \\ []) do
     max = Keyword.get(opts || [], :max_redirects, @max_redirects)
 
-    redirect(env, next, max)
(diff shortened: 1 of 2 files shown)


chulkilee:
See the discussion at #362

lib/tesla/middleware/follow_redirects.ex
```diff
@@ -26,7 +26,9 @@ defmodule Tesla.Middleware.FollowRedirects do
   def call(env, next, opts \\ []) do
     max = Keyword.get(opts || [], :max_redirects, @max_redirects)
 
-    redirect(env, next, max)
+    headers = Enum.filter(env.headers, &(not match?({"Authorization", _}, &1)))
```
chulkilee: All tesla adapters are downcasing the header key, so I believe this should check `authorization` not `Authorization`.

However, see the discussion at #362 for the scope.

teamon:
Let's continue in https://github.com/teamon/tesla/pull/371

PR #369, opened 2020-04-05
Add function to dynamically update existing client

dweinand:
When working with a library that uses Tesla with its own dynamic runtime like [Google's Gax](https://github.com/googleapis/elixir-google-api/blob/master/clients/gax/lib/google_api/gax/connection.ex#L52), there isn't an easy way if we also want to update the adaptor or append additional middleware.

With this PR one can do something like:

```elixir
Tesla.client([
  {Tesla.Middleware.Headers, [{"authorization", "Bearer #{token}"}
]})
|> Tesla.update_client([
  {Tesla.Middleware.Retry}
])
|> Tesla.get("/path")
```

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -499,6 +499,27 @@ defmodule Tesla do
   @spec client([Tesla.Client.middleware()], Tesla.Client.adapter()) :: Tesla.Client.t()
   def client(middleware, adapter \\ nil), do: Tesla.Builder.client(middleware, [], adapter)
 
+  @doc """
+  Dynamically update an existing client.
+
+  ```
+  # add dynamic middleware
+  client = %Tesla.Client{}
+  client = Tesla.update_client(client, [{Tesla.Middleware.Headers, [{"authorization", token}]}])
+  Tesla.get(client, "/path")
+
+  # configure adapter in runtime
+  client = %Tesla.Client{}
+  client = Tesla.update_client(client, [], Tesla.Adapter.Hackney)
+  client = Tesla.update_client(client, [], {Tesla.Adapter.Hackney, pool: :my_pool})
+  Tesla.get(client, "/path")
+  ```
+  """
+  @spec update_client(Tesla.Client.t(), [Tesla.Client.middleware()], Tesla.Client.adapter()) ::
+          Tesla.Client.t()
+  def update_client(client, middleware, adapter \\ nil),
+    do: Tesla.Builder.update_client(client, middleware, [], adapter)
+
   @deprecated "Use client/1 or client/2 instead"
   def build_client(pre, post \\ []), do: Tesla.Builder.client(pre, post)
 
--- a/lib/tesla/builder.ex
+++ b/lib/tesla/builder.ex
@@ -174,6 +174,21 @@ defmodule Tesla.Builder do
     %Tesla.Client{pre: runtime(pre), post: runtime(post), adapter: runtime(adapter)}
   end
 
+  def update_client(client, pre, post, adapter \\ nil)
+
+  def update_client(%Tesla.Client{} = client, pre, post, nil) do
+    %Tesla.Client{client | pre: client.pre ++ runtime(pre), post: client.post ++ runtime(post)}
+  end
+
+  def update_client(%Tesla.Client{} = client, pre, post, adapter) do
+    %Tesla.Client{
+      client
+      | pre: client.pre ++ runtime(pre),
+        post: client.post ++ runtime(post),
+        adapter: runtime(adapter)
+    }
+  end
+
   @default_opts []
 
   defp compile(nil), do: nil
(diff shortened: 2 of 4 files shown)


chulkilee:
`update_client` seems vague to me. For example, if new middlewares are given - will it *replace* the existing with it or *append* it to existing?

In the example 

```elixir
Tesla.client([
  {Tesla.Middleware.Headers, [{"authorization", "Bearer #{token}"}
]})
|> Tesla.update_client([
  {Tesla.Middleware.Retry}
])
|> Tesla.get("/path")
```

Why don't you build the middleware list "dynamically" and pass it later?

```elixir
[
  {Tesla.Middleware.Headers, [{"authorization", "Bearer #{token}"}
]
|> maybe_add_retry(other_args)
|> Tesla.client()
|> Tesla.get("/path")
```

----

I think it would be better if Tesla exposes 1) a function to return configured middlewares and 2) a function to replace existing middlewares with new one.

Use cases:

- replace header value of existing header value (e.g. renew) - without knowing all middlewares
- add/remove logging middleware of existing client

Note that all can be done by keeping all configuration given when building the client. However, that might not be a good idea since module creating the client and module using the client might be different..

However, the currently Tesla keeps expanded pre/post callbacks, with adapter middleware - so at least returning original middleware may not be trivial..

dweinand:
The idea Is that it would append middleware. If one wanted to replace them wholesale, one could simply create a new client.

It is true in this particular case, that Gax, is currently only adding the one simple middleware, which we could include ourselves. However, we’d then be breaking the contract and now would need to keep our code in sync with the internals of that lib.

It is also true, that we could try to update Gax to either use all the middleware and the adaptor we want (seems unlikely to scale) or to update its interface to take in user middleware that it combines with its own when creating the client. However, changing an existing widely used interface seems like a bigger ask than adding a new function here.

It is also true that we could just update the struct ourselves, but it’s internal structure is undocumented and the function that translates the arguments to `client` Into the structure is private.

Generally, I’m not aware of any reason why these need to all be setup in one go instead of providing the means to update after the fact.

One alternative could be to implement separate `update_adaptor` and `append_middleware` functions to make the intent clearer.

chulkilee:
I see the use case - I hit the similar case (e.g. I wanted to use one client for both oauth endpoint (need to use form) and other api endpoints (only using json).

### How to handle this with current Tesla

Since Tesla currently does not keep any state in Client or Env for connection, you can always keep and carry the whole information. Also you may consider introducing own client to make that easy.

```elixir
defmodule MyClient do
  defstruct[:middleware, :adapter]

  def post(client, path, body), do: Tesla.client(client.middleware, adapter) |> Tesla.post(path, body)
end
```

---

### What to "add"?

Please note that "update" is too vague, and a function to "append" is too specific. What if someone want to "prepend"? Then we need another function call. And not all functions in `post` or `pre` in client are middleware, I believe - so just prepending/appending to `pre`/`post` is not a correct solution.

Also for elixir convention

- "Update" usually means "replace" in Elixir world. 
- Usually it provides functions to update values, instead of helper functions (e.g. append/prepend)

- `Enum.update/4`
- `get_and_update/3`
- `Plug.put_*/3`

The tricky part is, as I said,`pre`/`post` functions != middlewares. They're even in different form. I understand "append_middleware" could be smallest and easiest change though :)

teamon:
I agree with @chulkilee that a generic `update` might be confusing. 
The internal structure of `Tesla.Client` is, and should remain private (it still carries the now-legacy `pre` and `post` parameters in a form that... is also a bit legacy due to macro-friendliness).

In one of my project I currently use this:

```elixir
@middleware [
  # ...
  Tesla.Middleware.Logger,
]

defp rpc_form(action, params) do
  middleware = [Tesla.Middleware.FormUrlencoded | @middleware]
  client = Tesla.client(middleware, @adapter)
  Tesla.post(client, action, params)
end

defp rpc_json(action, token, params) do
  middleware = [Tesla.Middleware.EncodeJson | @middleware]
  client = Tesla.client(middleware, @adapter)
  Tesla.post(client, action, params)
end
```

This is of course different from what you have - a `client` instead of a list of middlewares.

Since the internals of `Tesla.Client` are likely to change in the future, how about this:

```elixir
def my_client(gax_client)
  middleware = Tesla.Client.middleware(gax_client)
  adapter = Tesla.Client.adapter(gax_client)
  # modify middleware and adapter as needed
  Tesla.client(middleware, adapter)
end
```

Two new functions `Tesla.Client.middleware/1` and `Tesla.Client.adapter/1` that, given a client, would return the list of middlewares and an adapter in the same form as arguments to `Tesla.client/2`.

dweinand:
@teamon - That all makes sense and I'll close this PR out in favor of one that makes that change.

Given that `Tesla.Client` will now have functions and not just a struct definition, any objections to splitting it out into its own file and corresponding tests?

teamon:
Not at all, it's the right time to do it 👍

dweinand:
Closing in favor of #373

Issue #362, opened 2020-03-03
Leaking Secrets on Redirect!


lizziepaquette:
Same issue as myfreeweb/httpotion#135

Pull request here: https://github.com/teamon/tesla/pull/361

lizziepaquette:
@teamon I see you triaged, any thoughts on the fix?

chulkilee:
Where can we find the guideline for this? I don't think HTTP spec say anything about resending *any header*, and I guess there are "common" practice (e.g. by browser or popular http library).

Some notes:

- Why is only `authorization` header special? For example, there might be other headers with credentials
- What if when resending all header (including headers for credentials) is needed?
  - e.g. if the new location is relative path, or absolute but still in the same host (e.g. HATEOAS style)

teamon:
@lizziepaquette I am at the same point as @chulkilee here - it is trivial to implement it, but it might be very wrong. 

So far I think a reasonable option would be to:
- Add `:should_pass_auth_headers` option which defaults to `prev.host == next.host`
- Add `:pass_auth_headers` option which defaults to `["authorization"]`

I think with the above defaults we could treat this as a bug fix instead of a breaking change.

lizziepaquette:
While the HTTP spec doesn't have any explicit guidelines here I think it would be best to follow what curl does in this case: https://github.com/curl/curl/blob/6beb0eee/lib/http.c#L710

```
  /* To prevent the user+password to get sent to other than the original
     host due to a location-follow, we do some weirdo checks here */
  if(!data->state.this_is_a_follow ||
     conn->bits.netrc ||
     !data->state.first_host ||
     data->set.http_disable_hostname_check_before_authentication ||
     Curl_raw_equal(data->state.first_host, conn->host.name)) {
    result = output_auth_headers(conn, authhost, request, path, FALSE);
  }
  else
    authhost->done = TRUE;

  return result;
}
```
curl has a flag --location-trusted http://www.mit.edu/afs.new/sipb/user/ssen/src/curl-7.11.1/docs/curl.html#--location-trusted that allows the auth headers to be sent on redirects, but without that flag specified the auth headers are not sent if redirect's host is different from the first host. 

Right now auth headers are being sent through redirects without anyone expecting them to (as it is usually opt in). My pr stops the bleeding but does aggressively scrub the auth headers, not checking the host.

teamon:
Let's continue in https://github.com/teamon/tesla/pull/371

teamon:
[...]

PR #371, opened 2020-04-07
Keep authorization headers only if hostname is the same

teamon:
Replacement for https://github.com/teamon/tesla/pull/361, solving https://github.com/teamon/tesla/issues/362.

@lizziepaquette , @chulkilee Mind taking a look?

### DIFF
--- a/lib/tesla/middleware/follow_redirects.ex
+++ b/lib/tesla/middleware/follow_redirects.ex
@@ -54,6 +54,7 @@ defmodule Tesla.Middleware.FollowRedirects do
 
             env
             |> new_request(res.status, location)
+            |> filter_headers(env.url)
             |> redirect(next, left - 1)
         end
 
@@ -84,5 +85,15 @@ defmodule Tesla.Middleware.FollowRedirects do
     |> URI.parse()
     |> URI.merge(location)
     |> URI.to_string()
+  end
+
+  # See https://github.com/teamon/tesla/issues/362
+  @filter_headers ["authorization"]
+  defp filter_headers(env, orig_url) do
+    if URI.parse(env.url).host != URI.parse(orig_url).host do
+      %{env | headers: Enum.filter(env.headers, fn {k, _} -> k not in @filter_headers end)}
+    else
+      env
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/tesla/middleware/follow_redirects.ex
```diff
@@ -85,4 +86,14 @@ defmodule Tesla.Middleware.FollowRedirects do
     |> URI.merge(location)
     |> URI.to_string()
   end
+
+  # See https://github.com/teamon/tesla/issues/362
+  @filter_headers ["authorization"]
+  defp filter_headers(env, orig_url) do
```
chulkilee, line 92: ### cases

Several cases not handled

- header name in different case: e.g. both `authorization` `Authorization` should be handled. Maybe they're already lowercased before hitting here?
- same host but different port or scheme: e.g. `https://example.com` => `http://example.com:8080` - should drop the authorization

### userinfo in URL

Also.. is it possible that url in location (from redirect response) can have `userinfo`? If so, what should we do here?

1. request https://example.com with authorization
2. redirect to https://foo:bar@example.com/foo

In this case - should we drop original authorization *and* userinfo in the url?

### avoiding multiple URI.parse

Seems like it calls `URI.parse` somewhere - do you think we can refactor to avoid parsing multiple times?

We may make that as next refactoring PR - but just for the note...

test/tesla/middleware/follow_redirects_test.exs
```diff
@@ -187,4 +187,79 @@ defmodule Tesla.Middleware.FollowRedirectsTest do
...
+    end
+
+    setup :setup_client
+
+    test "Keep authorization header on redirect to the same domain", %{client: client} do
+      assert {:ok, env} =
+               Tesla.post(client, "http://example.com/keep", "",
+                 headers: [
+                   {"content-type", "text/plain"},
+                   {"authorization", "Basic: secret"}
```
chulkilee, line 227: The format of authorization header is without `:` and base64 encoded - so let's use the real value here

```
Basic Zm9vOmJhcg==
```

lib/tesla/middleware/follow_redirects.ex
teamon: > header name in different case
headers are downcased by adapters

> same host but different port or scheme 
✅ 

> avoiding multiple URI.parse
✅ 

> userinfo in URL
I _think_ that if the location header includes the userinfo we should not drop it. I'd leave it for now.

PR #378, opened 2020-04-29
Simplifying gun adapter logic

alex-strizhakov:
Changed:
- adapter would not open new connections on redirects, if passed `conn` option, because it doesn’t make sense if you use a connection pool and in some situations can lead to connections leaking. 
_For redirects handling with connection pool would be more correct to support redirects through custom `FollowRedirects` middleware, where you can properly open/close/checkin/checkout connections._
- little refactor
- some clean up in tests with removing duplicates


teamon:
👋 @alex-strizhakov is this one ready to be merged?

alex-strizhakov:
hey @teamon! yes, it's ready.

Issue #381, opened 2020-05-18
Finch adapter


teamon:
A tiny adapter for https://github.com/keathley/finch - a HTTP client with built-in pooling. 
Instead of building own mint+pooling adapter we could nicely integrate with Finch.

Something like this:

```elixir
defmodule Tesla.Adapter.Finch do
  @behaviour Tesla.Adapter

  def call(env, opts) do
    opts = Tesla.Adapter.opts(env, opts)    
    case Finch.request(name, env.method, env.url, env.headers, env.body, opts) do
      {:ok, ...} ->  # convert finch's response to %Tesla.Env{}
      {:error, ...} -> # ensure proper error handling
    end
  end
end
```

chulkilee:
Note

> Pools will be started for each configured `{scheme, host, port}` when Finch is started. For any unconfigured `{scheme, host, port}`, the pool will be started the first time it is requested. Note pools are not automatically terminated if they are unused, so Finch is best suited when you are requesting a known list of static hosts.

So I believe we need a way to trigger pruning unused pools (or a way to do it) from the library.

Monyk015:
Hey. I'm trying my hand at implementing this adapter and I'm not sure what to do with multipart and streaming. Finch's API doesn't seem to support it even though Mint does.

keathley:
@Monyk015 There is an open issue on finch to support a streaming API: https://github.com/keathley/finch/issues/58

keathley:
@Monyk015 streaming support has been added to finch: https://github.com/keathley/finch/pull/59

jrusso1020:
Hi @teamon my team needed to use a proxy with the Finch adapter, I wanted to pass in the `:proxy` and `:proxy_headers` to the pool based on this issue's examples https://github.com/keathley/finch/issues/85 like so
```
middleware = [...]

pool_opts = %{
  :default => [
    size: 5, 
    conn_opts: [
    proxy_headers: [{"proxy-authorization", "Basic " <> Base.encode64("#{proxy_user}:#{proxy_password}"}],
    proxy: {
      :http, 
      proxy_host, 
      proxy_port, 
      []
      }
    ]
  ]
}

adapter = {Tesla.Adapter.Finch, name: NamespaceFinch, pool: pool_opts }

Tesla.client(middleware, adapter)
```

But looking at the current Finch adapter it looks like we only expect three options

[...]

PR #402, opened 2020-07-13
Add Finch adapter

adriankumpf:
This adds an adapter for [Finch](https://github.com/keathley/finch) (#381).

Caveats:
- Since finch and its dependencies require Elixir 1.7 I removed those from the test matrix for now. I'm not sure if it's okay to drop support for v1.5 and v1.6 though. 
- For the "run tests against v1.0.0" CI job to succeed, the `:docs` option tests on the `v1.0.0-compat` branch must be updated.
- Request body streaming is not supported.


mix.exs
```diff
@@ -60,7 +60,8 @@ defmodule Tesla.Mixfile do
       {:ibrowse, "~> 4.4.0", optional: true},
       {:hackney, "~> 1.6", optional: true},
       {:gun, "~> 1.3", optional: true},
-      {:castore, "~> 0.1", optional: true},
+      {:finch, "~> 0.3", optional: true},
+      {:castore, "~> 0.1", optional: true, override: true},
```
teamon, line 64: why `override`?

test/tesla/adapter/finch_test.exs
```diff
@@ -0,0 +1,16 @@
+defmodule Tesla.Adapter.FinchTest do
+  use ExUnit.Case
+
+  @finch_name MyFinch
+
+  use Tesla.AdapterCase, adapter: Tesla.Adapter.Finch, adapter_opts: [name: @finch_name]
```
teamon, line 6: We could use the same syntax as in config:

```
use Tesla.AdapterCase, adapter: {Tesla.Adapter.Finch, [name: @finch_name]}
```

mix.exs
adriankumpf: Oh yeah, right 👍  Bumped it to v0.1.5 now since that is what Finch requires.

test/tesla/adapter/finch_test.exs
adriankumpf: Done!

teamon:
Merged as https://github.com/teamon/tesla/pull/420

PR #405, opened 2020-07-23
support function for BaseUrl

tobstarr:
For our use case we cannot use a static function for the BaseUrl but
rather read it from e.g. the ENV or the application config.

Also allowing a function for base gives us all these options.

### DIFF
--- a/lib/tesla/middleware/core.ex
+++ b/lib/tesla/middleware/core.ex
@@ -30,6 +30,10 @@ defmodule Tesla.Middleware.BaseUrl do
     |> Tesla.run(next)
   end
 
+  defp apply_base(env, base) when is_function(base) do
+    apply_base(env, base.(env))
+  end
+
   defp apply_base(env, base) do
     if Regex.match?(~r/^https?:\/\//i, env.url) do
       # skip if url is already with scheme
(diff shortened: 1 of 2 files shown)


teamon:
Hi,

If you need runtime configuration you can use [Runtime middleware](https://github.com/teamon/tesla#runtime-middleware).

If we allow function to be used as one middleware option it would need to be added to all middlewares and that increases the complexity. Dynamically creating the list of middlewares is the preferred way to go.

If you want to keep the `MyClient.some_fun()` calls you can do this by providing the default argument for client:

```ex
defmodule GitHub do
  # notice there is no `use Tesla`

  def user_repos(client \\ new(), login) do
    # pass `client` argument to `Tesla.get` function
    Tesla.get(client, "/user/" <> login <> "/repos")
  end

  def issues(client \\ new()) do
    Tesla.get(client, "/issues")
  end

  # build dynamic client based on runtime arguments
  def new() do
    url = Application.get_env(...)
    middleware = [
      {Tesla.Middleware.BaseUrl, url},
      Tesla.Middleware.JSON
    ]

    Tesla.client(middleware)
  end
end

tobstarr:
Thank you for the pointer. I just saw that Tesla.Middleware.Logger did support a function (but in that case only a function) as an option and thought this could be similar.

I thought about dynamically generating the client but thought this might a way to go.

PR #407, opened 2020-08-03
Improve retry middleware option parsing

britto:
When `delay` or `max_delay` are less than `1`, the Retry middleware currently crashes with the following error:

    ** (FunctionClauseError) no function clause matching in :rand.uniform_s/2
    (stdlib 3.13) rand.erl:326: :rand.uniform_s/2
    (stdlib 3.13) rand.erl:299: :rand.uniform/1
    (tesla 1.3.3) lib/tesla/middleware/retry.ex:106: Tesla.Middleware.Retry.backoff/3
    (tesla 1.3.3) lib/tesla/middleware/retry.ex:94: Tesla.Middleware.Retry.retry/3

It happens because `:rand.uniform/1` [requires](https://erlang.org/doc/man/rand.html#uniform-1) a positive integer.

### Original idea

~Make sure retry won't crash when either `delay` or `max_delay` are misconfigured.~

~This change adjusts `max_delay` to `1` when it isn't positive and also adjusts `delay` to make sure it falls within `1..max_delay`. Ideally, I believe we should will raise an `ArgumentError` right away when these values are invalid, e.g. when they are not integers, but that would be a bigger, breaking change. So this is a compromise we can safely make for now, without any behavior changes.~

### Update

This change applies strict validation to some retry middleware options. Now we explicitly fail when either one of the following misconfigurations are detected:
    
* `delay` is not a positive integer.
* `max_delay` is not a positive integer.
* `max_retries` is not a non-negative integer.

### DIFF
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -45,6 +45,9 @@ defmodule Tesla.Middleware.Retry do
   - `:should_retry` - function to determine if request should be retried
   """
 
+  # Not necessary in Elixir 1.10+
+  use Bitwise, skip_operators: true
+
   @behaviour Tesla.Middleware
 
   @defaults [
@@ -91,8 +94,8 @@ defmodule Tesla.Middleware.Retry do
 
   # Exponential backoff with jitter
   defp backoff(cap, base, attempt) do
-    factor = :math.pow(2, attempt)
-    max_sleep = trunc(min(cap, base * factor))
+    factor = Bitwise.bsl(1, attempt)
+    max_sleep = min(cap, base * factor)
     delay = :rand.uniform(max_sleep)
 
     :timer.sleep(delay)
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -39,9 +39,9 @@ defmodule Tesla.Middleware.Retry do
 
   ## Options
 
-  - `:delay` - The base delay in milliseconds (defaults to 50)
-  - `:max_retries` - maximum number of retries (defaults to 5)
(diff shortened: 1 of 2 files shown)


britto:
Hi, @teamon. Thanks for your input on this. I'll update this PR with the changes to make the middleware raise an error instead, hopefully with a helpful message, then I'll ask you to take another look as soon as it's done.

britto:
Updated and also seized the opportunity to validate `max_retries` as well. If it is accidentally set to a negative value, the middleware will keep retrying endlessly and that's probably an undesirable behavior. PTAL.

lib/tesla/middleware/retry.ex
```diff
@@ -39,12 +39,14 @@ defmodule Tesla.Middleware.Retry do
...
-  - `:delay` - The base delay in milliseconds (defaults to 50)
-  - `:max_retries` - maximum number of retries (defaults to 5)
-  - `:max_delay` - maximum delay in milliseconds (defaults to 5000)
+  - `:delay` - The base delay in milliseconds (positive integer, defaults to 50)
+  - `:max_retries` - maximum number of retries (non-negative integer, defaults to 5)
+  - `:max_delay` - maximum delay in milliseconds (positive integer, defaults to 5000)
   - `:should_retry` - function to determine if request should be retried
   """
 
+  use Bitwise, skip_operators: true
```
teamon, line 48: Is this `use` necessary?

britto: Good catch, it's no longer necessary. Fixed.

britto: Hmm, actually we do need it, for Elixir 1.9- support. I'll add it back.

https://github.com/elixir-lang/elixir/pull/9525

PR #417, opened 2020-09-07
Add basic Finch adapter

seancribbs:
Unfortunately, Finch does not support streaming request bodies directly, although
@keathley has assured me that adding it would not be difficult. This means that
we must consume the stream first before building the request.

Let me know anything else that needs to be added or accounted for!


seancribbs:
Woops, didn't see #402

jrusso1020:
@seancribbs I think this PR actually would be a good improvement to the Finch adapter because it would allow us to pass any options to it, whereas the current adapter only expects certain options

PR #420, opened 2020-09-10
Adriankumpf ak add finch adapter

teamon:


coladarci:
We are excited to try this out, any idea when a release will be cut for this? Thanks a ton for this!

teamon:
https://github.com/teamon/tesla/releases/tag/v1.4.0 sorry it took soooo long

adamu:
What's the advantage of this over the Mind adapter?

IslandUsurper:
One reason I've heard people recommend Finch over Tesla is that Tesla doesn't do connection pooling and makes a new connection for every request. I'm excited to see if this adapter will increase performance over the Mint adapter if you're going to do a lot of requests.

Issue #413, opened 2020-08-28
Logger Middleware does not have query variable


alexander7161:
When using the logger middleware and using https://hexdocs.pm/tesla/Tesla.Middleware.Logger.html#module-custom-log-format you should be able to print the query params.

Query should be added here:

https://github.com/teamon/tesla/blob/master/lib/tesla/middleware/logger.ex#L30

Ch4s3:
@teamon and/or @teamon I took a quick stab at this with this PR https://github.com/teamon/tesla/pull/422. If you have any feedback, that would be great!

alexander7161:
Thanks @Ch4s3 !

PR #422, opened 2020-09-11
Add private output/4 for :query

Ch4s3:
This allows logger middleware to add a formatting
key for the query string. It is implemented by
piping `env.query` to inspect/1 for the sake of
simplicity.

To address #413

### DIFF
--- a/lib/tesla/middleware/logger.ex
+++ b/lib/tesla/middleware/logger.ex
@@ -5,7 +5,7 @@ defmodule Tesla.Middleware.Logger.Formatter do
   # https://github.com/elixir-lang/elixir/blob/v1.6.4/lib/logger/lib/logger/formatter.ex
 
   @default_format "$method $url -> $status ($time ms)"
-  @keys ~w(method url status time)
+  @keys ~w(method url status time query)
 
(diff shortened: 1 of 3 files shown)


lib/tesla/middleware/logger.ex
```diff
@@ -27,6 +27,7 @@ defmodule Tesla.Middleware.Logger.Formatter do
     Enum.map(format, &output(&1, request, response, time))
   end
 
+  defp output(:query, env, _, _), do: env.query |> inspect()
```
teamon, line 30: We should output the same format that will actually be used by tesla - https://github.com/teamon/tesla/blob/master/lib/tesla.ex#L500

Ch4s3: I considered that, and wasn't sure if it made sense to make `encode_query/1` available in `Tesla.Middleware.Logger.Formatter`.

I'm happy to make that change. I'll take a second look shortly.

Ch4s3: Ok, I updated the PR. Let me know if there's anything else I should do.

PR #423, opened 2020-09-17
Fix wrong output in README.md

jechol:
`url` doesn't include query parameters.


teamon:
Nice catch!

I think we can remove this example, it doesn't bring much value.

jechol:
@teamon Removed the line instead of fixing as you said.

PR #424, opened 2020-09-30
Gun adapter proxy auths

matijavibe:
This PR aims to address the lack of ability to use Basic auth when using proxy (regular or Socks) via GUN adapter.
Also a pattern match is added to not force TLS in case proxy is used, and leave it to GUN to use TLS if the proxy port is 443, otherwise TCP will be used. 
The combination of TLS and a port other than 443 for proxy can be accomplished by passing the transport option when creating the Tesla Client and adding it as Adapter opt.


teamon:
@alex-strizhakov 👀 🙏

teamon:
@alex-strizhakov Could you take a look at this PR?

lib/tesla/adapter/gun.ex
```diff
@@ -38,6 +38,8 @@ if Code.ensure_loaded?(:gun) do
     - `:close_conn` - Close connection or not after receiving full response body. Is used for reusing gun connections. Defaults to `true`.
     - `:certificates_verification` - Add SSL certificates verification. [erlang-certifi](https://github.com/certifi/erlang-certifi) [ssl_verify_fun.erl](https://github.com/deadtrickster/ssl_verify_fun.erl)
     - `:proxy` - Proxy for requests. **Socks proxy are supported only for gun master branch**. Examples: `{'localhost', 1234}`, `{{127, 0, 0, 1}, 1234}`, `{:socks5, 'localhost', 1234}`.
+                NOTE: By default GUN uses TLS as transport if the specified port is 443, if TLS is required for proxy conection on another port please specify transport using the Gun options below otherwise tcp will be used
```
alex-strizhakov, line 41:
```suggestion
                NOTE: By default GUN uses TLS as transport if the specified port is 443, if TLS is required for proxy connection on another port please specify transport using the Gun options below otherwise tcp will be used
```

lib/tesla/adapter/gun.ex
```diff
@@ -345,6 +366,18 @@ if Code.ensure_loaded?(:gun) do
 
     defp tunnel_tls_opts(opts, _, _), do: opts
 
+    defp add_proxy_auth_credentials(opts, %{proxy_auth: {username, password}})
+         when not is_nil(username) and not is_nil(password),
```
alex-strizhakov, line 370: maybe `is_binary/1` guard will be more suitable here for checking username and password?

alex-strizhakov (changes requested):
the rest looks good 👍

lib/tesla/adapter/gun.ex
matijavibe: You are absolutely right, corrected.

lib/tesla/adapter/gun.ex
matijavibe: Nice catch, corrected :)

alex-strizhakov (approved):

PR #434, opened 2020-11-26
Fix type error

MasashiYokota:
I've fixed a type error on `Tesla.MultiPart.add_file/3`.

I've simply removed its option because there is no `:read` option for `File.stream!/3`.
If this patch has any problems. please let me know.

Thank you!

### DIFF
--- a/lib/tesla/multipart.ex
+++ b/lib/tesla/multipart.ex
@@ -103,7 +103,7 @@ defmodule Tesla.Multipart do
         false -> headers
       end
 
-    data = File.stream!(path, [:read], 2048)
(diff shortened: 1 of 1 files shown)


tanguilp:
Thank you @MasashiYokota

This incorrect parameter causes Dialyzer to emit warnings in projects using Tesla. Surprisingly, it doesn;t when running it on the tesla library alone.

`:read` isn't a valid option indeed for streaming files, as specified here: https://hexdocs.pm/elixir/1.11.3/File.html#stream!/3

teamon:
Thanks!

PR #436, opened 2020-11-27
Allow setting logger's format at runtime

blisscs:
Hi, 

Currently, I am having an issue that I can't set the format of the logger at runtime because the current implementation takes the format string from the module attribute which is set at compile time. With this pull request, we will be able to set the format at run time with the below code sample.

```elixir
  middleware = [
    {Tesla.Middleware.Logger, format: "$method $url -> $status ($time ms)"}
    # ,... more middlewares
  ]

  client = Tesla.client(middleware)
```

### DIFF
--- a/lib/tesla/middleware/logger.ex
+++ b/lib/tesla/middleware/logger.ex
@@ -161,7 +161,8 @@ defmodule Tesla.Middleware.Logger do
     config = Keyword.merge(@config, opts)
 
     format =
-      Keyword.get(config, :format)
+      config
+      |> Keyword.get(:format)
       |> Formatter.compile()
 
     level = log_level(response, config)
--- a/lib/tesla/middleware/logger.ex
+++ b/lib/tesla/middleware/logger.ex
@@ -149,7 +149,6 @@ defmodule Tesla.Middleware.Logger do
   alias Tesla.Middleware.Logger.Formatter
 
   @config Application.get_env(:tesla, __MODULE__, [])
-  @format Formatter.compile(@config[:format])
 
   @type log_level :: :info | :warn | :error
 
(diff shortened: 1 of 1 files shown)


blisscs:
Hi @teamon,

I have been using tesla for some time now. I have a little issues in that i can't configure my logger's format at runtime. This merge request will help to resolve that. please help review this merge request, if you have any comments please let's me know

Thanks.
Suracheth

blisscs:
Hi @teamon Please let's me know what do you think of this merge request to allow setting the format of logger at runtime.

lib/tesla/middleware/logger.ex
```diff
@@ -161,8 +160,12 @@ defmodule Tesla.Middleware.Logger do
 
     config = Keyword.merge(@config, opts)
 
+    format =
+      Keyword.get(config, :format)
```
thiamsantos, line 164:
```suggestion
      config
      |> Keyword.get(:format)
```
:nail_care:

blisscs:
@thiamsantos I update the code to use pipe. Let's me know if you need other changes.

lib/tesla/middleware/logger.ex
```diff
@@ -161,8 +160,13 @@ defmodule Tesla.Middleware.Logger do
 
     config = Keyword.merge(@config, opts)
 
+    format =
+      config
+      |> Keyword.get(:format)
+      |> Formatter.compile()
+
```
teamon, line 167: I think we should do something like this:

```elixir
format = if opts[:format], do: Formatter.compile(opts[:format]), else: @format
```

so in case of not using runtime config we still benefit from compile-time format compilation.

blisscs: @teamon 

I updated the code to use compile-time format. when there is no runtime format given in the option.
but we can't use `opts[:format]` since it's keyword not a map. So will have to use `Keyword.get/2` instead

teamon:
Perfect!

PR #440, opened 2021-01-14
Use original HTTP method to retrieve digest authorization variables

### LINKED ISSUE #439: Use original HTTP method to retrieve digest authorization variables
When using `Tesla.Middleware.DigestAuth`, Tesla always performs HTTP `GET` request to retrieve authentication variables:
https://github.com/teamon/tesla/blob/586c54372cef5ff8bcb570d2c1ad06fa2b81b833/lib/tesla/middleware/digest_auth.ex#L57-L63

Example code:
```elixir
defmodule Demo do
  use Tesla

  plug Tesla.Middleware.DigestAuth, username: "guest", password: "guest"
  plug Tesla.Middleware.Logger, debug: :false

  def run() do
    delete("http://jigsaw.w3.org/HTTP/Digest/")
  end
end

Demo.run()
```

Output:
```
10:18:03.149 [error] GET http://jigsaw.w3.org/HTTP/Digest/ -> 401 (303.257 ms)

10:18:03.311 [error] DELETE http://jigsaw.w3.org/HTTP/Digest/ -> 401 (137.208 ms)
```

Please note that when authenticated, the response status should be `405 Method Not Allowed` (because `DELETE` is not supported by this endpoint), not `401 Unauthorized`.

However, servers may not include `WWW-Authenticate` header in the response when HTTP `GET` request method is used. For instance, [sabre-io/Baikal](https://github.com/sabre-io/Baikal) does not return `WWW-Authenticate` header when `GET` method is used instead of `MKCALENDAR`. In such case Tesla cannot compose `Authorization` header for subsequent request and silently ignores it.

Therefore original HTTP method should be used to retrieve authentication variables. This behaviour has been verified with Python Requests library as well as cURL on digest authentication demo available at http://jigsaw.w3.org/HTTP/Digest/. 
```sh
curl -v "http://jigsaw.w3.org/HTTP/Digest/" --digest -u guest:guest -X DELETE
```
Notice that the first request uses `DELETE` method, not `GET`:
```
*   Trying 128.30.52.21...
* TCP_NODELAY set
* Connected to jigsaw.w3.org (128.30.52.21) port 80 (#0)
* Server auth using Digest with user 'guest'
> DELETE /HTTP/Digest/ HTTP/1.1
> Host: jigsaw.w3.org
> User-Agent: curl/7.64.1
> Accept: */*
> 
< HTTP/1.1 401 Unauthorized
< date: Thu, 14 Jan 2021 10:06:04 GMT
< content-length: 261
< content-type: text/html;charset=ISO-8859-1
< server: Jigsaw/2.3.0-beta2
< www-authenticate: Digest realm="test", domain="/HTTP/Digest", nonce="d1c4f5824ff8e0254beed4fb6eab353d"
< 
* Ignoring the response-body
* Connection #0 to host jigsaw.w3.org left intact
* Issue another request to this URL: 'http://jigsaw.w3.org/HTTP/Digest/'
* Found bundle for host jigsaw.w3.org: 0x7ff773716790 [can pipeline]
* Could pipeline, but not asked to!
* Re-using existing connection! (#0) with host jigsaw.w3.org
* Connected to jigsaw.w3.org (128.30.52.21) port 80 (#0)
* Server auth using Digest with user 'guest'
> DELETE /HTTP/Digest/ HTTP/1.1
> Host: jigsaw.w3.org
> Authorization: Digest username="guest", realm="test", nonce="d1c4f5824ff8e0254beed4fb6eab353d", uri="/HTTP/Digest/", response="e4d1252a5cfc737e69b859e848bb064c"
> User-Agent: curl/7.64.1
> Accept: */*
> 
< HTTP/1.1 405 Method Not Allowed
< date: Thu, 14 Jan 2021 10:06:04 GMT
< allow: HEAD,GET,OPTIONS,TRACE
< content-length: 26
< content-type: text/html
< server: Jigsaw/2.3.0-beta2
< 
* Connection #0 to host jigsaw.w3.org left intact
Method DELETE not allowed.* Closing connection 0
```

tomekzaw:
Fixes #439

### DIFF
--- a/lib/tesla/middleware/digest_auth.ex
+++ b/lib/tesla/middleware/digest_auth.ex
@@ -58,7 +58,7 @@ defmodule Tesla.Middleware.DigestAuth do
     with {:ok, unauthorized_response} <-
            env.__module__.request(
              env.__client__,
-             method: env.method,
+             method: env.opts[:pre_auth_method] || env.method,
              url: env.url,
              opts: Keyword.put(env.opts || [], :digest_auth_handshake, true)
            ) do
--- a/lib/tesla/middleware/digest_auth.ex
+++ b/lib/tesla/middleware/digest_auth.ex
@@ -56,9 +56,10 @@ defmodule Tesla.Middleware.DigestAuth do
 
   defp authorization_vars(env, opts) do
     with {:ok, unauthorized_response} <-
-           env.__module__.get(
+           env.__module__.request(
              env.__client__,
-             env.url,
+             method: env.method,
+             url: env.url,
              opts: Keyword.put(env.opts || [], :digest_auth_handshake, true)
            ) do
       {:ok,


teamon:
@hodak Would you mind taking a look?

hodak:
Looks legit 👍

PR #442, opened 2021-01-19
Introduce semantic versioning

KrzysiekJ:
This is a proposition to document what apparently already takes place. Also Hackney example version is adjusted to be consistent in both examples and Jason example version is tightened to not allow `>= 2.0`.


teamon:
Rebased and merged as https://github.com/teamon/tesla/commit/13e58ac2ab743aa0de8e583d0bd36324af4beed4

Thanks!

PR #446, opened 2021-01-30
Misc doc changes

kianmeng:
List of changes:

- Bump license year
- Use common source url
- Fix typos
- Fix markdown
- Add release changelog to readme
- Add hexdocs badge
- Use and set ex_doc to latest version
- Update gitignore

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,7 @@
 
 [![Build Status](https://github.com/teamon/tesla/workflows/Test/badge.svg?branch=master)](https://github.com/teamon/tesla/actions)
 [![Hex.pm](https://img.shields.io/hexpm/v/tesla.svg)](http://hex.pm/packages/tesla)
+[![Hexdocs.pm](https://img.shields.io/badge/hex-docs-lightgreen.svg)](https://hexdocs.pm/tesla/)
 [![Hex.pm](https://img.shields.io/hexpm/dt/tesla.svg)](https://hex.pm/packages/tesla)
 [![Hex.pm](https://img.shields.io/hexpm/dw/tesla.svg)](https://hex.pm/packages/tesla)
 [![codecov](https://codecov.io/gh/teamon/tesla/branch/master/graph/badge.svg)](https://codecov.io/gh/teamon/tesla)
@@ -13,6 +14,8 @@ It embraces the concept of middleware when processing the request/response cycle
 > Note that this README refers to the `master` branch of Tesla, not the latest
   released version on Hex. See [the documentation](http://hexdocs.pm/tesla) for
   the documentation of the version you're using.
+
+For list of changes, checkout the latest [release notes](https://github.com/teamon/tesla/releases).
 
 ---
 
@@ -59,18 +62,14 @@ See below for documentation.
 
 ## Installation
 
-Add `tesla` as dependency in `mix.exs`:
+Add `:tesla` as dependency in `mix.exs`:
 
 ```elixir
 defp deps do
   [
     {:tesla, "~> 1.4.0"},
-
-    # optional, but recommended adapter
-    {:hackney, "~> 1.16.0"},
-
-    # optional, required by JSON middleware
-    {:jason, ">= 1.0.0"}
+    {:hackney, "~> 1.16.0"}, # optional, but recommended adapter
+    {:jason, ">= 1.0.0"} # optional, required by JSON middleware
   ]
 end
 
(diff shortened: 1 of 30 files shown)


README.md
```diff
@@ -139,7 +138,7 @@ This is very similar to how [Plug Router](https://github.com/elixir-plug/plug#th
 
 ## Runtime middleware
 
-All HTTP functions (`get`, `post`, etc.) can take a dynamic client as the first argument.
+All HTTP functions (`get`, `post`, and etc) can take a dynamic client as the first argument.
```
chulkilee, line 141: Incorrect. `a, b, etc.` is the correct form (no `and`, and `etc.` not `etc` since `.` is for abbreviation)

https://www.grammarly.com/blog/et-cetera-etc/
https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Punctuation/faq0040.html

Probably we may rewrite this like this:

> All HTTP functions, such as `Tesla.get/3` and `Tesla.post/4`, can take a ...

lib/tesla/middleware/keep_request.ex
```diff
@@ -1,6 +1,6 @@
 defmodule Tesla.Middleware.KeepRequest do
   @moduledoc """
-  Store request body & headers into opts.
+  Store request `body` and `headers` into opts.
```
chulkilee, line 3: I think "body" and "headers" here are referring the HTTP terms, not fields in the struct. However, `opts` is a field.

What about this? 

> Keeps request body and headers in `opts`.

lib/tesla/middleware/logger.ex
```diff
@@ -40,9 +40,11 @@ defmodule Tesla.Middleware.Logger do
   @moduledoc """
   Log requests using Elixir's Logger.
 
-  With the default settings it logs request method, url, response status and time taken in milliseconds.
+  With the default settings it logs request method, url, response status and
```
chulkilee, line 43: url => URL, if we want to make it consistent with other parts

lib/tesla/middleware/method_override.ex
```diff
@@ -1,9 +1,9 @@
 defmodule Tesla.Middleware.MethodOverride do
   @moduledoc """
-  Middleware that adds X-Http-Method-Override header with original request
+  Middleware that adds `X-Http-Method-Override` header with original request
```
chulkilee, line 3: I think it would be better to downcase the example as in the code.

Suggestion

> Changes the request method to POST and adds `x-http-method-override` header with the original request method

README.md
kianmeng: @chulkilee You're right! Appreciate it.

chulkilee (changes requested):
:heart: 

I added some feedback.

And general comments

- I don't think adding a line break for long lines add significant values... 
- Probably better to rephrase the headline of moduledoc to what it does, starting with third-person singular verbs for consistency (within tesla, and it's common in Elixir itself) - I added one example for a middleware. It is really useful when seeing a list of modules. e.g. `Use ...` => `Uses ...`

README.md
```diff
@@ -228,7 +229,7 @@ def new(...) do
 end
 ```
 
-Passing directly to `get`/`post`/etc.
+Passing directly to `get` or etc.
```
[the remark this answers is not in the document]
kianmeng: Since the examples after the statement have two different function calls, I've updated to:

```
  232 Passing directly to request functions such as `MyClient.get/3` or `Tesla.get/3`.                      
  233                                                                                                       
  234 ```elixir                                                                                             
  235 MyClient.get("/", opts: [adapter: [recv_timeout: 30_000]])                                            
  236 Tesla.get(client, "/", opts: [adapter: [recv_timeout: 30_000]])                                       
  237 ```  
```

lib/tesla/middleware/compression.ex
```diff
@@ -2,9 +2,9 @@ defmodule Tesla.Middleware.Compression do
   @moduledoc """
   Compress requests and decompress responses.
 
-  Supports "gzip" and "deflate" encodings using erlang's built-in `:zlib` module.
+  Supports `"gzip"` and `"deflate"` encodings using Erlang's built-in `:zlib` module.
```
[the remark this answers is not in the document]
kianmeng: Reverted to "gzip" and "deflate" instead.

lib/tesla/middleware/method_override.ex
kianmeng: Based on https://opensocial.github.io/spec/2.5.1/Core-API-Server.xml#rfc.section.2.1.1.1 via https://en.wikipedia.org/wiki/List_of_HTTP_header_fields,
`X-HTTP-Method-Override` seems to be the right name with case.

teamon:
Rebased and merged in https://github.com/teamon/tesla/commit/65d178ae4b754b7afb8791e2f83ca689604345e5

Thanks!

PR #454, opened 2021-03-16
feat: add new event metadata to tesla telemetry

polvalente:
This PR aims to provide a way for relaying through the `tesla.request.stop` event the unresolved path when using both Telemetry and PathParams middlewares.

The issue with the current implementation is that in success cases, it's impossible to retrieve the unresolved path from the event metadata

### DIFF
--- a/lib/tesla/middleware/keep_request.ex
+++ b/lib/tesla/middleware/keep_request.ex
@@ -1,6 +1,6 @@
 defmodule Tesla.Middleware.KeepRequest do
   @moduledoc """
-  Store request body & headers into opts.
+  Store request body, headers and path into `%Tesla.Env{}` opts.
 
   ## Example
 
@@ -9,9 +9,10 @@ defmodule Tesla.Middleware.KeepRequest do
     use Tesla
 
     plug Tesla.Middleware.KeepRequest
+    plug Tesla.Middleware.PathParams
   end
 
-  {:ok, env} = MyClient.post("/", "request-data")
+  {:ok, env} = MyClient.post("/users/:user_id", "request-data", opts: [path_params: [user_id: "1234]])
 
   env.body
   # => "response-data"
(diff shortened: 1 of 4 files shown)


teamon:
I see. Assuming you use `Telemetry` middleware _before_ `PathParams` I think the best way would be to use something similar to `KeepRequest`:

```elixir
defmodule KeepRequestURL do
  def call(env, next, _opts) do
    env
    |> Tesla.put_opt(:req_url, env.url)
    |> Tesla.run(next)
  end
end

# then in client
plug Tesla.Middleware.Telemetry
plug Tesla.Middleware.KeepRequestURL
plug Tesla.Middleware.PathParams
```

Then in you telemetry handlers you should be able to access the original URL via `env.opts[:req_url]`.

I'd be happy to accept a PR adding `:req_url` to the built-in `KeepRequest` middleware too.

polvalente:
> I see. Assuming you use `Telemetry` middleware _before_ `PathParams` I think the best way would be to use something similar to `KeepRequest`:
> 
> ```elixir
> defmodule KeepRequestURL do
>   def call(env, next, _opts) do
>     env
>     |> Tesla.put_opt(:req_url, env.url)
>     |> Tesla.run(next)
>   end
> end
> 
> # then in client
> plug Tesla.Middleware.Telemetry
> plug Tesla.Middleware.KeepRequestURL
> plug Tesla.Middleware.PathParams
> ```
> 
> Then in you telemetry handlers you should be able to access the original URL via `env.opts[:req_url]`.
> 
> I'd be happy to accept a PR adding `:req_url` to the built-in `KeepRequest` middleware too.

Yeah, the `Telemetry` middleware is before `PathParams`, but without this fix the URL ends up being resolved for happy-path scenarios, and thus this PR.

I'll change this PR to adding `:req_url` to the `KeepRequest` middleware then, a I'll also take the time to add an example to the `Telemetry` docs.

Thanks for the review and the suggestion!

polvalente:
@teamon the PR has now been updated as per your suggestion. I took the liberty to add an "integration" test for KeepRequest + PathParams + Telemetry, as well as an example for other `Tesla.Middleware.Telemetry` users.

If there's a better place to add these, please let me know!

teamon:
Nice! 😍

PR #456, opened 2021-03-18
Fix race condition of Tesla.Mock.mock_global/1

sircinek:
Problem:
Intermittent test failures when using `Tesla.Mock.mock_global/1` 
```elixir
      ** (exit) exited in: GenServer.call(#PID<0.16741.0>, {:update, #Function<2.71048325/1 in Tesla.Mock.agent_set/1>}, 5000)
          ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
      code: Tesla.Mock.mock_global(fn
      stacktrace:
        (elixir 1.10.4) lib/gen_server.ex:1023: GenServer.call/3
```

Current implementation:
`agent_set/1` is not atomic, and there's no guarantee pid returned by `Process.whereis/1` will be alive when making `Agent.update/2` call. That's because `Agent` process is linked to the test process, but the `exit(:shutdown)` signal is asynchronous, thus the same process can "leak" between test cases in certain edge cases.

Proposed implementation:
Instead of using plain `Agent.start_link/2` use `ExUnit.Callbacks.start_supervised!/1`, that puts the `Tesla.Mock` agent under the test supervision tree and guarantees the process is terminated synchronously, thus prevents the leak between test cases.

### DIFF
--- a/lib/tesla/mock.ex
+++ b/lib/tesla/mock.ex
@@ -218,8 +218,14 @@ defmodule Tesla.Mock do
 
   defp agent_set(fun) do
     case Process.whereis(__MODULE__) do
-      nil -> Agent.start_link(fn -> fun end, name: __MODULE__)
-      pid -> Agent.update(pid, fn _ -> fun end)
+      nil ->
+        ExUnit.Callbacks.start_supervised!(%{
+          id: __MODULE__,
+          start: {Agent, :start_link, [fn -> fun end, [{:name, __MODULE__}]]}
+        })
+
+      pid ->
+        Agent.update(pid, fn _ -> fun end)
     end
   end
 


teamon:
Thanks!!

ZeLarpMaster:
Hey there, I was using `mock_global` in my `test_helper.exs` because I had to mock a call done by my application on startup (and before tests are executed), but this change broke my tests because `start_supervised/2 can only be invoked from the test process` and because `test_helper.exs` isn't a test process. Maybe what I was doing wasn't supported or a good idea, but I was wondering on what's your opinion on that and if you have suggestions on how to solve this issue? Afaik there isn't another `mock_global` which could be used globally outside tests. Maybe a different `mock_global` scoped to tests would be better than changing `mock_global` to be scoped to tests? Not sure on what's your opinion on this.

If you need access, [this is the project where I do that](https://github.com/ZeLarpMaster/Mobius) and I noticed the issue because of [this PR by renovate](https://github.com/ZeLarpMaster/Mobius/pull/84)

teamon:
@ZeLarpMaster As a quick answer, you can work around this by replacing:

```elixir
Tesla.Mock.mock_global(myfun)
``` 

with

```elixir
{:ok, _} = Agent.start_link(myfun, name: Tesla.Mock)
```

PR #457, opened 2021-03-22
Add documentation for `options/0` type and exclude when `docs: false`

seancribbs:
Previously, in modules that `use Tesla, docs: false` the `option/0` type would
display in the generated documentation without any functions that refer to it in
the documentation.


teamon:
Thanks!

PR #462, opened 2021-04-24
Add BearerAuth middleware

mdlkxzmcp:
Adds a `Tesla.Middleware.BearerAuth` middleware.

Basically, I had to use the bearer auth method, but Tesla doesn't have a middleware for it. So here is my small contribution :)

Wanted to include a wiki article on it in the module docs, but it's only listed as a part of Oauth...

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,7 @@ This is very similar to how [Plug Router](https://github.com/elixir-plug/plug#th
 ### Auth
 
 - [`Tesla.Middleware.BasicAuth`](https://hexdocs.pm/tesla/Tesla.Middleware.BasicAuth.html) - HTTP Basic Auth
+- [`Tesla.Middleware.BearerAuth`](https://hexdocs.pm/tesla/Tesla.Middleware.BearerAuth.html) - HTTP Bearer Auth
 - [`Tesla.Middleware.DigestAuth`](https://hexdocs.pm/tesla/Tesla.Middleware.DigestAuth.html) - Digest access authentication
 
 ### Error handling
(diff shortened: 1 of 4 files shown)


teamon:
Hi @mdlkxzmcp

I appreciate the contribution! 

If I understand correctly, what this middlewares does when used like this:

```elixir
Tesla.client [
  {Tesla.Middleware.BearerAuth, Map.merge(%{token: token}, opts))}
]
```

is essentially the same as using `Tesla.Middleware.Headers` directly:

```elixir
Tesla.client [
  {Tesla.Middleware.Headers, [{"authorization", "Bearer #{token}}]}
]
```

In general, adding such new middleware shouldn't be needed (as one can use the more generic `Headers` middleware).
However, we do include `BaseAuth` middleware for a nicer user experience, and I think it doesn't hurt to have a `BearerAuth` too.

Having said that, I'd like to add as little code as possible, as each line of code needs to be maintained for a long time. And this applies also to tests.

I'd suggest a minimal implementation, something along (not tested)

```elixir
def call(env, next, opts) do
  token = Keyword.fetch!(opts, :token)

  env
  |> Tesla.put_headers({"authorization", "Bearer #{token}})
  |> Tesla.run(next)
end
```

and then just one or two test cases [testing the middleware in isolation](https://github.com/teamon/tesla/blob/master/test/tesla/middleware/header_test.exs).

Please do let me know if you have any questions regarding all that.

mdlkxzmcp:
Hey @teamon! 

It is indeed essentially the same as setting the header, however, I felt that, since it's related to authentication, it would be beneficial to have it as its own middleware. 

I'm totally ok with rewriting it to be simpler like you showed. I simply wanted to follow the logic from other middleware modules :)

Should have a rewrite by the end of the day :>

PR #463, opened 2021-04-24
Use :telemetry.attach/4 in examples.

maartenvanvliet:
Arity 3 does not exist here, the fourth argument is the handler config https://hexdocs.pm/telemetry/index.html that can be nil.

In the examples an anonymous function is passed in. For performance reasons these should be avoided. So perhaps that should be changed as well?


teamon:
Thanks,I'd keep the examples minimal as they are now.

PR #465, opened 2021-05-17
Unify configuration

### LINKED ISSUE #234: Unify configuration
Currently there is some inconsistency regarding adapter/middleware configuration.
For example, logger format can be configured with `config :tesla, Tesla.Middleware.Logger, debug: false` but `log_level` can't. 

This all should be unified with clear rules of precedence.
Also relates to #219

thiamsantos:
Implementation of  #234

This is a draft of the implementation, I'm opening to check if the approach makes sense. cc: @teamon 

## Proposed solution

Use [nimble_options](https://hex.pm/packages/nimble_options) to validate the opts given to a middleware. For a option to be fetched from Application Configuration, it needs to set `global: true` in the `config_schema` of the middleware.

It follows a clear precedence of first fetching from application and then overriding with the configuration given directly as options to the middleware. It also always fetches the configuration in runtime.

Based on the schema a documentation for the options is generated:

Screenshot_2021-05-17 Tesla Middleware Logger — tesla v1 4 1

## TODO

- [ ] Documentation
- [ ] Unit Tests
- [ ] Use in all middlewares

### DIFF
--- a/lib/tesla/middleware/config.ex
+++ b/lib/tesla/middleware/config.ex
@@ -1,0 +1,58 @@
+defmodule Tesla.Middleware.Config do
+  @moduledoc false
+
+  def build!(middleware, schema, opts) do
+    global_keys =
+      schema
+      |> Enum.filter(fn {_k, v} -> v[:global] == true end)
+      |> Enum.map(fn {k, _v} -> k end)
+
+    :tesla
+    |> Application.get_env(middleware, [])
+    |> Keyword.take(global_keys)
+    |> Keyword.merge(opts)
+    |> NimbleOptions.validate(to_nimble(schema))
+    |> handle_validate(middleware)
+  end
+
+  def docs(schema) do
+    schema
+    |> to_nimble()
+    |> NimbleOptions.docs()
+  end
+
+  defp to_nimble(schema) do
+    Enum.map(schema, fn {k, v} ->
+      {global, value} = Keyword.pop(v, :global, false)
+
+      if global == true do
+        {k,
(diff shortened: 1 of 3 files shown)


teamon:
Thank you, @thiamsantos, for picking this up!

I think this looks good, especially that it would introduce proper options validation instead of ad-hoc solutions implemented in some middlewares.  I'm naturally hesitant with introducing a new dependency, but since it will be used by all middlewares it makes sense to include it (instead of reimplement option validation inside tesla).

In general, there are two cases that need to be taken care of:
1. Client module in the same codebase - full access to code, configuration can be done via options passed directly to middleware, application env useful when providing different setting for different environments (dev/test/prod)
2. Client module from third-party library - no direct access to code, configuration only possible if exposed or via application env

Recently, I've been using this pattern to configure first-party clients:

```elixir
defmodule SomeAPI do
  def get_foo(client \\ new(), id), do: ...
  
  def new(opts \\ []) do
    config = Application.get_env(:myapp, __MODULE__, [])
    opts = Keyword.merge(config, opts)
    
    middleware = [
      {Tesla.Middleware.BaseUrl, opts[:base_url]},
     ...
    ]

    Tesla.client(middleware, config[:adapter])
  end
end
```

It is especially useful for changing base URL and/or adapter in test.


@hodak @amatalai @rafapaez Would you mind taking a look at this proposal in terms of how would it affect your codebases?

teamon:
cc @elixir-tesla/tesla

yordis:
I am gonna come back to this at some point, I like the idea

PR #473, opened 2021-07-01
Document that Tesla.Mock can be configured to return error tuples

bschmeck:
I wasn't able to find how to do this in the documentation, but it's supported by the `call` function.  This PR adds an example to the `Setting up mocks` section of the docs.

I'm happy to rephrase things if "error tuple" is not the right term to use.


teamon:
Thanks!

In fact, `Tesla.Mock` accepts multiple values for convenience:

https://github.com/teamon/tesla/blob/3549695c79d8aa77c734353bf502599c25ef4828/lib/tesla/mock.ex#L197-L211

Issue #470, opened 2021-06-23
Reusing Mint connections from the adapter itself


jotakami:
I'm trying to implement some basic connection pooling using the Mint adapter, and I thought it would be super easy to just reuse the connections that the adapter opens rather than rewriting all of that logic. The adapter does a good job of hiding the connection though, and it seems like the only way to expose it is to use the `body_as: :chunks` option and extract it from the `body` map. It would be nice if the connection could be exposed with the `:plain` and `:stream` options as well, at least when the `close_conn: false` option is set.

However, there is another issue I'm running into, which is that `close_conn: true` is _always_ set, overriding the existing setting, whenever the adapter opens the connection itself. This occurs on line 159 of `mint.ex`, which is preceded by a comment stating "If there were redirects, and passed `closed_conn: false`, we need to close opened connections to these intermediate hosts." As far as I can tell, the option is overwritten regardless of whether there were redirects or not.

I can work around this by manually setting the option back to `false` when I get the response from the adapter, but this is not very elegant and it seems like a bug that the option is overwritten no matter what.

PR #477, opened 2021-07-13
Force close connection only on redirects in Mint adapter

jotakami:
Currently the `Mint` adapter will force the `close_conn` option to `true` anytime the connection is opened by the adapter itself. Judging by the comment in the source, it seems that this is really only supposed to occur when there is a host mismatch with the provided `conn` (e.g. due to using `Middleware.FollowRedirects`).

It's worth noting that, as far as I can tell, the builtin adapters do not follow consistent behavior in this situation (host mismatch). The `Mint` adapter discards the provided connection and opens a new one, while the `Gun` adapter appears to just return an error.


jotakami:
Partially addresses #470

PR #479, opened 2021-07-14
Document the plug behavior of runtime function evaluation

PragTob:
First of all, thanks a lot for Tesla - it's a joy to work with! 💚 

This confused us a lot when working with Tesla. The docs don't
seem to mention but it does work (we are using it and tried
out it's behavior, the function are definitely called at run
time and well the implementation uses `Macro.escape`).

As some feedback, as mentioned this behavior was irritating to us. I do this here vs. a separate issue as I believe it's intended and as we at version 1.x+ breaking compatibility would of course be bad. However, this design is confusing imo because:

* it works differently from the known `plug` usage
* looking at the code, it looks like it'd be run at compile time and hence be wrong, "magic" makes it not be wrong
* if this is intended behvaior, I feel like passing in the reference to a function (`&get_api_key/0`) would be a better way to accomplish this vs. calling the function but not really now `get_api_key()`

Anyhow, thanks a lot for all of your work, you're really helping the elixir community 💚 🎉 🙌 

IMG_20180518_161925_01

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,8 @@ client |> GitHub.user_repos("teamon")
 client |> GitHub.get("/me")
 ```
 
+However, if insted of a pure String you use a function call that is evaluated at run time and works: `plug Tesla.Middleware.Headers, [{"authorization", get_api_key()}]`
+
 ## Adapters
 
 Tesla supports multiple HTTP adapter that do the actual HTTP request processing.
--- a/lib/tesla/builder.ex
+++ b/lib/tesla/builder.ex
@@ -84,6 +84,11 @@ defmodule Tesla.Builder do
   @doc """
   Attach middleware to your API client.
 
+  Notice that this has one important difference to Plug's plug that you might be used to.
+  Here `opts` aren't passed by value but as the AST and evaluated at run time.
+  This means, while you can't pass a secret string directly, you can call a function that
+  retrieves the secret and it will be called at run time.
+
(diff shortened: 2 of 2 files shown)


teamon:
Before I get to write a proper comment, here’s the tldr: 

Passing functions to “plug”-ged middleware should not work, as it’s too confusing. If you need to pass runtime values to middleware you should be using Runtime Middleware as described in the README. In fact, that whole macro-based style should be removed, or at least be a very specific case of runtime approach. But, of course, backwards compatibility. 

There is a separate issue about making runtime middleware the default in README, unfortunately I haven’t got enough spare cycles to properly take care of that.

PragTob:
@teamon we're in agreement then (see my comments in the description about how confusing this is). I take it then, you'd rather not have this documented to not spread its usage? Happy to close then.

Usually happy to help, but due to arm problems sadly won't be able to.

JesseHerrick:
@teamon @PragTob I'm happy to take a stab at updating the README to emphasize runtime usage over macro usage.

teamon:
@JesseHerrick 👋

teamon:
(oh boy it's been a while...)

@PragTob I'm going to close this one for the reasons you mentioned  🐰 💜 

@JesseHerrick Would you like to tackle https://github.com/elixir-tesla/tesla/issues/377 ?

PR #485, opened 2021-08-06
Compression Middleware: Add accept-encoding header when appropriate

wkirschbaum:
Avoid having to add an extra header when using the Compression or Compression derivatives middleware.


wkirschbaum:
It seems like identity is implied, so don't think it adds value to add it to the list, but not sure.

teamon:
@amatalai Would you mind taking a look?

teamon:
Thanks!

PR #486, opened 2021-08-17
Support nested list of tuples in query parameters

ejscunha:
# Description
Besides supporting nested keyword lists for query parameters, now a list of tuples where the first element of the tuple is binary is also supported.

This way there is no need to transform the list of tuples into a keyword, where the casting of keys into atoms could be dangerous when user input is used.

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -537,7 +537,7 @@ defmodule Tesla do
 
   @doc false
   def encode_pair({key, value}) when is_list(value) do
-    if Keyword.keyword?(value) do
+    if Keyword.keyword?(value) or Enum.all?(value, &is_list_of_tuples?/1) do
       Enum.flat_map(value, fn {k, v} -> encode_pair({"#{key}[#{k}]", v}) end)
     else
       Enum.map(value, fn e -> {"#{key}[]", e} end)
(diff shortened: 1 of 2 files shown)


lib/tesla.ex
```diff
@@ -546,4 +546,7 @@ defmodule Tesla do
 
   @doc false
   def encode_pair({key, value}), do: [{key, value}]
+
+  defp is_list_of_tuples?({k, _}) when is_binary(k), do: true
+  defp is_list_of_tuples?(_), do: false
 end
```
teamon, line 552: Given how [`Keyword.keyword?`](https://github.com/elixir-lang/elixir/blob/v1.12.2/lib/elixir/lib/keyword.ex#L127) is implemented we could do the same:

```elixir
# on the top
if list_of_tuples?(value) do

# below

def list_of_tuples?([{k, _} | rest]) when is_atom(k) or is_binary(k), do: list_of_tuples?(rest)
def list_of_tuples?([]), do: true
def list_of_tuples?(_other), do: false
``

ejscunha: Thanks, it looks much better.

teamon (approved):

PR #489, opened 2021-09-20
Allow mime ~> 2.0

maennchen:
The major change in `mime` is that there are less supported mime types: https://github.com/elixir-plug/mime/blob/master/CHANGELOG.md

I don't think that should be a problem for this library though.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -52,7 +52,7 @@ defmodule Tesla.Mixfile do
 
   defp deps do
     [
-      {:mime, "~> 1.0"},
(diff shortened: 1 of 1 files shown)


maennchen:
Oh, this would also mean dropping older elixir versions than 1.10. So not sure if you want to move ahead.

teamon:
Can we relax the dependency, but keep 1.x in mix.lock? 🤔

maennchen:
@teamon Sure, done

teamon:
Thank you! 💜

Issue #487, opened 2021-08-30
OpenApi Generator


teamon:
## What is this all about?

Generate full tesla-based API client based on Open Api specification file, like this:

```ex
defmodule Petstore do
  use Tesla.OpenApi, spec: "petstore.json"
end
```

[Tweet](https://twitter.com/iteamon/status/1432361728029888522?s=20) that started it all.

## Why?

We already have the great [open_api_spex](https://github.com/open-api-spex/open_api_spex) library to generate API spec from Phoenix apps. We should also have an easy way to use this spec from another Elixir app, as an API client.

## Why not use existing OpenAPI generators?

Existing generators have quite big dependencies (like docker or java). In case of Elixir, they generate a separate mix package with many files. This is suboptimal, as for every spec change one must regenerate and release a new package.

Here, I propose an alternative, where there are no client API packages at all. The client application can simply depend on tesla and define its own API client module, with customized adapters and middleware. The client is generated on the fly, during regular compilation.

## Draft To-Do list
- [ ] Generate schemas (models) with `decode(body) :: {:ok, schema} | {:error, ...}` including validation
    Very basic version done without error handling nor validation
- [ ] Generate operations with & without body & query params
    Very basic version done (no body)
- [ ] Generate typespecs & docs for operations
    Very basic version done (no body)
- [ ] Figure out how to handle authentication, i.e. how to use Tesla auth middlewares and pass the _token_ taking into account multiple authorization schemas
- [ ] Keep the `client` as `Tesla.client()` (unless it is super hard, then maybe it should be changed)
- [ ] Allow adding more middleware (logging, telemetry, ...)
- [ ] Allow choosing adapter
- [ ] Test with [all examples](https://github.com/OAI/OpenAPI-Specification/tree/main/examples)
- [ ] Test with big specs, like [Slack](https://api.slack.com/specs/openapi/v2/slack_web.json)

maennchen:
@teamon I was trying to achieve something similar (it is by no means finished). Maybe you'll find some inspiration?

https://github.com/jshmrtn/openapi-compiler

yordis:
@teamon how are you thinking to deal with `$ref` and remote `$ref` in schemas and multiple files to define the spec?

PR #490, opened 2021-09-23
OpenAPI Generator

teamon:
see #487


teamon:
Moved into separate repo https://github.com/elixir-tesla/tesla_openapi

PR #498, opened 2021-10-29
Add stream_response option to hackney adapter

### LINKED ISSUE #271: Streaming Response
According to the README, Tesla supports streaming if the underlying adapter supports it.

From the code and the examples I found it seems though that streaming means in this instance, streaming of the request.

I am interested in how I would be able to stream the response of a request (very useful for something like large downloads).

Is this supported, if so: how?

tim-smart:
Playing around with streaming responses (#271).

Only implemented for hackney here.

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -50,7 +50,8 @@ defmodule Tesla.Env do
           status: status,
           opts: opts,
           __module__: atom,
-          __client__: client
+          __client__: client,
+          __pid__: pid()
         }
 
   defstruct method: nil,
@@ -61,7 +62,8 @@ defmodule Tesla.Env do
             status: nil,
             opts: [],
             __module__: nil,
-            __client__: nil
+            __client__: nil,
+            __pid__: nil
 end
 
 defmodule Tesla.Middleware do
@@ -322,7 +324,7 @@ defmodule Tesla do
   end
 
   defp prepare(module, %{pre: pre, post: post} = client, options) do
-    env = struct(Env, options ++ [__module__: module, __client__: client])
+    env = struct(Env, options ++ [__module__: module, __client__: client, __pid__: self()])
     stack = pre ++ module.__middleware__ ++ post ++ [effective_adapter(module, client)]
     {env, stack}
   end
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -32,6 +32,7 @@ defmodule Tesla.Env do
   @type param :: binary | [{binary | atom, param}]
   @type query :: [{binary | atom, param}]
   @type headers :: [{binary, binary}]
+  @type response :: :default | :stream
 
   @type body :: any
   @type status :: integer | nil
@@ -48,6 +49,7 @@ defmodule Tesla.Env do
           headers: headers,
           body: body,
           status: status,
+          response: response,
           opts: opts,
           __module__: atom,
           __client__: client,
@@ -60,6 +62,7 @@ defmodule Tesla.Env do
             headers: [],
             body: nil,
             status: nil,
+            response: :default,
             opts: [],
             __module__: nil,
             __client__: nil,
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -53,41 +53,42 @@ if Code.ensure_loaded?(:hackney) do
         Tesla.build_url(env.url, env.query),
         env.headers,
         env.body,
+        env.response,
         Tesla.Adapter.opts(env, opts)
         |> Keyword.put_new(:stream_owner, env.__pid__)
       )
     end
 
-    defp request(method, url, headers, %Stream{} = body, opts),
-      do: request_stream(method, url, headers, body, opts)
+    defp request(method, url, headers, %Stream{} = body, response, opts),
+      do: request_stream(method, url, headers, body, response, opts)
 
-    defp request(method, url, headers, body, opts) when is_function(body),
-      do: request_stream(method, url, headers, body, opts)
+    defp request(method, url, headers, body, response, opts) when is_function(body),
+      do: request_stream(method, url, headers, body, response, opts)
 
-    defp request(method, url, headers, %Multipart{} = mp, opts) do
+    defp request(method, url, headers, %Multipart{} = mp, response, opts) do
       headers = headers ++ Multipart.headers(mp)
       body = Multipart.body(mp)
 
-      request(method, url, headers, body, opts)
+      request(method, url, headers, body, response, opts)
     end
 
-    defp request(method, url, headers, body, opts) do
+    defp request(method, url, headers, body, type, opts) do
       response = :hackney.request(method, url, headers, body || '', opts)
-
-      case {Keyword.get(opts, :stream), Keyword.get(opts, :stream_owner)} do
-        {true, pid} when is_pid(pid) -> handle_stream(response, pid)
-        _ -> handle(response)
-      end
+      handle(response, type, Keyword.get(opts, :stream_owner))
     end
 
-    defp request_stream(method, url, headers, body, opts) do
+    defp request_stream(method, url, headers, body, type, opts) do
+      handle_wrap = fn result ->
+        handle(result, type, Keyword.get(opts, :stream_owner))
+      end
+
       with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
         case send_stream(ref, body) do
-          :ok -> handle(:hackney.start_response(ref))
-          error -> handle(error)
+          :ok -> handle_wrap.(:hackney.start_response(ref))
+          error -> handle_wrap.(error)
         end
       else
-        e -> handle(e)
+        e -> handle_wrap.(e)
       end
     end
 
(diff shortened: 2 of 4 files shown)


test/tesla/adapter/hackney_test.exs
```diff
@@ -62,4 +62,17 @@ defmodule Tesla.Adapter.HackneyTest do
...
     assert {:error, :fake_error} = call(request)
   end
+
+  test "get with `stream_to_pid: pid` option" do
+    request = %Env{
+      method: :get,
+      url: "#{@http}/ip"
+    }
+
+    assert {:ok, %Env{} = response} = call(request, stream_to_pid: self())
```
teamon, line 72: When passing pid explicitly, I'd expect to handle incoming message manually. 
Is there anything stopping us from using `steam_response: true` instead?

Does the implementation handle multiple concurrent requests originating from the same process?

teamon:
Amazing! Let's get the response streaming become a thing 👍

test/tesla/adapter/hackney_test.exs
tim-smart: Hackney requires you to transfer ownership of the response to the desired PID, otherwise it will get GCed.

Workaround will be to capture `self()` during the request phase, and assume that will be the PID that will consume the steam.

maxmarcon:
👍  for this!

test/tesla/adapter/hackney_test.exs
teamon: And what happens when the stream is consumed from a different process?

tim-smart: From what I recall, Hackney shuts down the request after a short amount of time.

So for long running responses you get a partial reply.

cgarvis:
@tim-smart @teamon anything I can do to get this over the finish line?  This would allow me to  remove a dependency on Downstream from my app.

Hajto:
Do you need any help with this? I would also want to help to get this through the finish line.

byronalley:
I'll add my voice to the "I'd love to see this feature merged, can I help?" crowd!

tim-smart:
Added some changes to improve the API surface.

* Added `response` to the `Env` struct. Defaults to `:default`, but also has a `:stream` option
* Added `__pid__` to `Env`, for capturing the pid of the requestor process
* Added `stream_owner` to the hackney adapter options, which defaults to `__pid__` from the `Env`

New API example:

```elixir
{:ok, %Env{body: stream}} = Tesla.get("https://google.com", response: :stream)

# Use the response stream
data = Enum.join(stream)
```

tim-smart:
Maybe leaving this as an adapter option is better, as it will not be implemented for every adapter.

connorjacobsen:
Hey @teamon I used this as a base for a version that is adapter option driver: https://github.com/connorjacobsen/tesla/tree/hackney-response-streaming

I would love to be able to use hackney to stream responses like you and Tim have detailed here. Is there anything I can do to help make that happen? Happy to take a stab at some more code as needed, would just want some direction from you so it fits with where you want to guide the library.

connorjacobsen:
@tim-smart @yordis is there anything I can do to help with this PR? Or the fork I linked above? tesla is a fantastic library and I would love to be able to contribute a little something back to it if it's helpful!

yordis:
@connorjacobsen, by all means, please take the lead.

In the worst case, @tim-smart solved it differently, and we commit to one of the solutions, and you get to learn about Tesla internals; from my perspective, still a huge win.
In the best case, we fix the issue.

I am unfamiliar with `hackney,` but I can help to some extent.

connorjacobsen:
@yordis I am equally happy with either solution. I basically just took @tim-smart's [comment](https://github.com/elixir-tesla/tesla/pull/498#issuecomment-1276898245) and riffed on it in a way that kept mostly everything isolated to the Hackney adapter itself rather than adding the `response` field to the core `Tesla.Env` struct.

So I guess my question for you + others is: do you want to isolate this to the Hackney adapter _or_ should we go down a more generalized path adding streaming for all adapters (all seem to support streaming responses). If the latter, do we want to pursue that one-by-one or all at once? Do you have a preference?

As a meta question: is this the right place to discuss this? Is there somewhere better?

yordis:
Ideally, we add streaming to all adapters.

About one-by-one vs. all at once. If you are giving it a try, I rather you do a big PR where you experiment and find a path forward that works for the official adapters, and if we need to, we split the PR into smaller ones if they are too big.
My worry is to merge something that prevents me from releasing a version, or something broken or error-prune.

teamon:
There is also this PR https://github.com/elixir-tesla/tesla/pull/540 that implements streaming for Finch. 

I’d say we go with the `response: :stream` opt and implement each adapter separately.

yordis:
Either way is fine by me

noozo:
Any progress on this?

feld:
Is there a branch that best embodies an approach that is likely to be merged that can be collaborated on and/or is there a way we can put a bounty on this?

yordis:
@feld I haven't used Hackney for years by now, so I wouldn't consider myself qualified to have strong professional judgment.

I would appreciate it if folks test it and share your code review.

yordis:
@tomekowal do you think you could help me with this one? Should we wait for v4? My concern is that since, Erlang the pkg is global, most likely folks are stuck in older version.

lib/tesla/adapter/hackney.ex
```diff
@@ -95,21 +111,55 @@ if Code.ensure_loaded?(:hackney) do
...
+          fn _ ->
+            case :hackney.stream_body(ref) do
+              :done ->
+                {:halt, nil}
+
+              {:ok, data} ->
+                {[data], nil}
+
+              {:error, reason} ->
+                raise inspect(reason)
```
tomekowal, line 145: I am pretty sure, there could be a way to communicate the error without raising in the middle of the stream.

lib/tesla.ex
```diff
@@ -1,60 +1,308 @@
...
+          query: query,
+          url: url,
+          headers: headers,
+          body: body,
+          status: status,
+          response: response,
+          opts: opts,
+          __module__: atom,
+          __client__: client,
+          __pid__: pid()
```
tomekowal, line 56: I am not sure I like adding option to the entire Env that is used by just one adapter.

tomekowal:
@yordis I've checked the PR and I don't fully like it. I think we can avoid adding `__pid__` to env. Also, streaming changed heavily in Hackney 4.x. In 1.x it was tied to TCP connection (hence the `controlling_process`). In 4.x the architecture is different and doesn't require knowing such internals.

This means, that if we make 1.x streaming work, we will need a separate implementation for 4.x.

My recommendation would be to give up on streaming in 1.x, upgrade to 4.x and only implement streaming with that one (so we live with one implementation).

Let's keep this PR open until I am done with 4.x upgrade and when that lands, let's close it.

yordis:
Hey folks, I am not sure if we should make `:hackney` to work or eventually, we will get a `gen_http` in OTP itself, or simply people should use `Mint` or `Mint-based` adapter instead of `:hackney`.

Personally speaking, I haven't use `:hackney` directly for more than half a decade right now, so I am not qualify to understand all the internals of it and daily, have direct experience.

This PR has been opened for a very long time by now, so, I think it is safe to assume that not that many people care as much.

Therefore, I am gonna close it. If you would like to get this into `master` I will help either way, please open a new PR.

---

Hey @tomekowal in the end, you are effectively the only contributor or user that interact here around it, if you need the changes, I am here to help. Up to you, no expectations, thank you so much for the support with it.

tomekowal:
I don't believe Tesla needs that at all. It is possible to pass `async: true` like this:

```
  describe "async response streaming" do
    test "echoed POST body reassembles from `:hackney_response` chunks" do
      request = %Env{
        method: :post,
        url: "#{@http}/post",
        headers: [{"content-type", "text/plain"}],
        body: "hello streamed world"
      }

      assert {:ok, %Env{status: 200, body: ref}} = call(request, async: true)
      assert is_reference(ref) or is_pid(ref)
      assert drain_async_body(ref) =~ "hello streamed world"
    end

    test "GET /stream/N delivers all chunks via async messages" do
      request = %Env{
        method: :get,
        url: "#{@http}/stream/5"
      }

      assert {:ok, %Env{status: 200, body: ref} = response} = call(request, async: true)
      assert is_reference(ref) or is_pid(ref)
      assert Tesla.get_header(response, "content-type") =~ "application/json"

      body = drain_async_body(ref)

      ids =
        ~r/"id":(\d+)/
        |> Regex.scan(body)
        |> Enum.map(fn [_, id] -> String.to_integer(id) end)

      assert ids == [0, 1, 2, 3, 4]
    end
  end

  defp drain_async_body(ref, acc \\ []) do
    receive do
      {:hackney_response, ^ref, :done} ->
        IO.iodata_to_binary(Enum.reverse(acc))

      {:hackney_response, ^ref, chunk} when is_binary(chunk) ->
        drain_async_body(ref, [chunk | acc])
    after
      5_000 -> flunk("timed out waiting for async body chunk")
    end
  end
```

If somebody uses Hackney's streaming, it is probably because they want to receive those messages to a particular process. I am not sure it makes sense to wrap it in streaming.
Personally, I can live without it.

Issue #367, opened 2020-03-25
Adapter options aren't merged correctly


23shortstop:
I need to use a proxy server for requests for some services. I created a reusable module for this:

```
defmodule Tesla.Proxy do
  defmacro __using__(_opts) do
    quote do
      @proxy_url Application.fetch_env!(:app, :proxy_url)
      @proxy_user Application.fetch_env!(:app, :proxy_user)
      @proxy_password Application.fetch_env!(:app, :proxy_password)

      plug Tesla.Middleware.Opts,
        adapter: [proxy: @proxy_url, proxy_auth: {@proxy_user, @proxy_password}]
    end
  end
end
```

Then I used this module in an API client module where I also needed to add another adapter options:

```
defmodule App.ApiClient do
  use Tesla
  use Tesla.Proxy

  plug Tesla.Middleware.Opts, adapter: [ssl_options: [{:versions, [:tlsv1]}]]

  ...
end
```

I believe that Tesla should merge all adapter options, but now it's not.
At first all options are appended to one list:
```
opts = [
  adapter: [ssl_options: [{:versions, [:tlsv1]}]],
  adapter: [proxy: ..., proxy_auth: {..., ...}]
]

```
Then adapter options retrieved using square brackets here:
https://github.com/teamon/tesla/blob/d5ce3a75d7ff285563fdfd47905e66ba4a0433a3/lib/tesla.ex#L215-L220
And we get only the first value in case there are few pairs with the same key in a Keyword.

To solve the issue I suggest to rewrite this function this way:
```
  @spec opts(Keyword.t(), Tesla.Env.t(), Keyword.t()) :: Keyword.t()
  def opts(defaults \\ [], env, opts) do
    defaults
    |> Keyword.merge(opts || [])
    |> Keyword.merge(env_opts(env))
  end

  defp env_opts(env) do
    env.opts
    |> Keyword.get_values(:adapter)
    |> List.flatten
  end
```

teamon:
I would discourage usage of macros for DRY. It's better to use the [Runtime middleware](https://github.com/teamon/tesla#runtime-middleware).

I think `Tesla.Middleware.Opts` should use `Keyword.merge`, instead of simple `++`, but unfortunately it would be a breaking change. 

On the other hand, currently it _doesn't work_, and it seems like there is a little chance someone is relying on this behaviour.

sambhavsaxena:
I would like to work on this issue

teamon:
[...]

PR #500, opened 2021-11-07
fix: merge adapter opts

brenomaia:
A simple way to resolve issue #367 , hope I understood it correctly.


lib/tesla/middleware/core.ex
```diff
@@ -125,6 +125,10 @@ defmodule Tesla.Middleware.Opts do
   @behaviour Tesla.Middleware
 
   @impl Tesla.Middleware
+  def call(env, next, adapter: adapter_opts) do
```
teamon, line 128: This pattern will only match `[adapter: x]` or `[adapter: x, other: opt]` but not `[other: ops, adapter: x]`.

teamon (changes requested):

teamon:
@brenomaia 👋 Could you push the requested changes so we can merge this PR? :)

yordis:
I couldn't modify this PR 😭 so I had to create #613

PR #503, opened 2021-11-29
fix: jitter should user * instead of +

polvalente:
This PR is fairly straight-forward. The current implementation basically has no jitter (which isn't necessarily a serious bug, but is a bug still).

### DIFF
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -103,7 +103,7 @@ defmodule Tesla.Middleware.Retry do
 
     # This ensures that the delay's order of magnitude is kept intact, while still having some jitter.
     # Generates a value x where 1-jitter_factor <= x <= 1 + jitter_factor
-    jitter = 1 + 2 * jitter_factor * :rand.uniform() - jitter_factor
+    jitter = 1 - jitter_factor * :rand.uniform()
 
     # The actual delay is in the range max_sleep * (1 - jitter_factor), max_sleep * (1 + jitter_factor)
     delay = trunc(max_sleep * jitter)
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -106,7 +106,7 @@ defmodule Tesla.Middleware.Retry do
     jitter = 1 + 2 * jitter_factor * :rand.uniform() - jitter_factor
 
     # The actual delay is in the range max_sleep * (1 - jitter_factor), max_sleep * (1 + jitter_factor)
-    delay = trunc(max_sleep + jitter)
+    delay = trunc(max_sleep * jitter)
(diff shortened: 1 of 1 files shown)


teamon:
The current implementation is definitely wrong, running 20 attempts with default parameters produces something like this:

```
50
101
200
401
801
1600
3200
5001
5000
5001
5001
5001
5000
5001
5000
5001
5000
5000
5001
5001
5000
```

Not much jitter there.

With this PR the values look like this:

```
58
105
160
476
660
1555
2680
5485
5930
5615
5676
4813
5112
5424
4684
4780
5936
4052
5182
4001
5972
```

Jitter is much better, but it goes way too high over `max_delay`.

To fix this we'd need to change the jitter calculation to:


```ex
jitter = 1 - jitter_factor * :rand.uniform()
```

which will result in this values

```
45
99
179
373
705
1399
2891
4326
4678
4097
4242
4540
4118
4040
4716
4321
4387
4531
4535
4558
4766
```

teamon:
@polvalente Would you like to update your PR to include the jitter calculation change?

lib/tesla/middleware/retry.ex
```diff
@@ -106,7 +106,7 @@ defmodule Tesla.Middleware.Retry do
     jitter = 1 + 2 * jitter_factor * :rand.uniform() - jitter_factor
 
     # The actual delay is in the range max_sleep * (1 - jitter_factor), max_sleep * (1 + jitter_factor)
-    delay = trunc(max_sleep + jitter)
+    delay = trunc(max_sleep * jitter)
```
polvalente, line 109: @teamon What do you think about changing it to:

`delay = trunc(base * factor * jitter) |> min(cap)`

this way we get a hard limit on `cap`, but still have jitter around the target value whenever possible.

It's basically the same code as we have here but with the `min(cap, _)` moved to the end of the calculation

polvalente: If you prefer your suggestion anyway, just let me know and I'll update accordingly

polvalente: Actually, we'd need to still calculate the `max_sleep` first anyway. I'm going to use your suggestion which keeps the code simpler

teamon: Since the base grows exponentially, you'll quickly end up with cap without jitter:

```
49
94
168
353
841
1747
3625
5000
5000
5000
5000
5000
5000
5000
5000
5000
5000
5000
5000
5000
5000
```

polvalente: I've submitted your suggestion :)

polvalente: Oh wait, gotta update the comment above it :)

PR #504, opened 2021-12-06
docs: enhance the Logger middleware document.

### LINKED ISSUE #379: Documentation Issues: Logger Middleware
I was running it to a situation where I set the log_level of a Logger middleware to always be `:debug`, but I was still seeing log output with the Logger log_level set to `warn`.

Looking at the code, it does something slightly different than was the documentation says.

If the request returns {:error, whatever}, the log level is always set to :error, the custom log level function is never called. The custom log level is only used when an {:ok, ...} is returned.

Yarnus:
Hey @teamon , I'm trying to resolve #379, hope I understood it correctly. 

But this case confused me:

>If the request returns {:error, whatever}, the log level is always set to :error, the custom log level function is never called. The custom log level is only used when an {:ok, ...} is returned.

Which level of priority should be higher ?

- scenario A: let user's option prior to `log_level/2` whenever.
- scenario B: only apply log_level option when `{:ok, env}`. (**I took this one ATM.** )

I'm still a green hand for contribution. Please do not hesitate to give your comment or suggestion. 

😄 I welcome any questions or comments you might have. (even coding style or variable name.)

### DIFF
--- a/lib/tesla/middleware/logger.ex
+++ b/lib/tesla/middleware/logger.ex
@@ -54,7 +54,7 @@ defmodule Tesla.Middleware.Logger do
 
   ## Options
 
-  - `:log_level` - custom function for calculating log level (see below)
+  - `:log_level` - custom option for calculating log level when get `{:ok, response}` after call.(see below)
   - `:filter_headers` - sanitizes sensitive headers before logging in debug mode (see below)
   - `:debug` - show detailed request/response logging
 
@@ -81,7 +81,7 @@ defmodule Tesla.Middleware.Logger do
   - `:warn` - for 3xx responses
   - `:info` - for 2xx responses
 
-  You can customize this setting by providing your own `log_level/1` function:
+  You can customize this setting by providing your own `log_level/1` function when you get `{:ok, response}` after call:
 
   ```
   defmodule MyClient do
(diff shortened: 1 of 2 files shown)


teamon:
Hey!

Unfortunately changing the contract of `log_level` function from `(env) -> atom` to `({:ok, env} | {:error, any}) -> atom` would be a breaking change, so we can't do that.

In general, I think it makes sense to have log level function accept both ok & error cases.

One way to solve this, would be to add a new option, i.e. `level: atom | fun` and deprecate the current `log_level` option with some logic like this:

```ex
cond do
  log_level option given and level option given -> 
    error (can't provide both)
  log_level option given ->
    show deprecation warning
    use this function: fn {:ok, env} -> log_level.(env); _ -> :error end
  level option given ->
    use level
  else ->
    use default level function
end
```

Yarnus:
Okay, I’m gonna try and fix it

yordis:
What should we do about this PR and https://github.com/elixir-tesla/tesla/issues/379 issue?

teamon:
Ideally I'd go with https://github.com/elixir-tesla/tesla/pull/504#issuecomment-996699875 (new option, deprecate old one) and update docs accordingly.

PR #508, opened 2021-12-17
HTTP Cache Middleware

teamon:
Continuation of https://github.com/teamon/tesla/pull/264

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -6,12 +6,17 @@ export_locals_without_parens = [
   adapter: 2
 ]
 
+internal = [
+  assert_cached: 1,
+  refute_cached: 1
+]
+
 [
   inputs: [
     "lib/**/*.{ex,exs}",
     "test/**/*.{ex,exs}",
     "mix.exs"
   ],
-  locals_without_parens: export_locals_without_parens,
+  locals_without_parens: export_locals_without_parens ++ internal,
   export: [locals_without_parens: export_locals_without_parens]
 ]
--- a/lib/tesla/middleware/cache.ex
+++ b/lib/tesla/middleware/cache.ex
@@ -106,56 +106,36 @@ defmodule Tesla.Middleware.Cache do
       end
     end
 
-    defp ttl({env, cc}) do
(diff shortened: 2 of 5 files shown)


teamon:
To answer the [last comment from previous PR](https://github.com/teamon/tesla/pull/264#issuecomment-962813743)

> I've also made a try on store_opts, as a mean to have "multi-tenant" caches, i.e., ServiceA and ServiceB each have their own cache. The implementation is the simplest as possible (passing tuples around), and I can refactor if we decide that's a good approach API-wise.

I think it would be best to have separate cache for each client as a default.

> Talking about API, I'm not sure if I really like the result. The Store behaviour feels a bit inflated, with a total of 4 args for put for example. One alternative I can think of is, instead of a module, store argument is actually a name which we can use via some Registry - that feels more "Elixir-y". What I don't like about that is that feels overcomplicated for Tesla.

One option would be to not ship any storage adapter at all. Even the simples one (ets) gets tricky prevtty quickly when you start to think of concurrent access, expirations etc (that's why we have con_cache). 

Maybe instead we should shit the HTTP Cache logic only and provide guidance on how to use X as a cache storage (con_cache, redix, ecto, ...)

tanguilp:
FYI I'm working on an [HTTP caching library in Erlang](https://github.com/tanguilp/http_cache) with some niceties such as handling of revalidation, auto-compression, handling the whole RFC7234 (e.g. deleting cache when `DELETE`ing) for private of shared proxies, and more.

I'm also working on an [Erlang LRU backend](https://github.com/tanguilp/http_cache_store_native) for this, that handles invalidation by URI or by alternate key, clustering, etc.

In addition to an HTTP caching Plug, I'm planning to write a Tesla middleware for this :) Not writing here to brag about it, but to avoid duplicate work depending on what your goals are with this PR.

teamon:
@andrewhr @tanguilp So... how can we move forward with this? :)

tanguilp:
FYI I've released [`tesla_http_cache`](https://hex.pm/packages/tesla_http_cache) a few days ago.

PR #511, opened 2022-01-03
Add a middleware to encode/decode MessagePack payloads

chaodhib:
Heavily based on the Json Middleware. Uses https://github.com/lexmag/msgpax as the engine.


lib/tesla/middleware/message_pack.ex
```diff
@@ -0,0 +1,159 @@
...
+  - `:decode` - decoding function
+  - `:encode` - encoding function
+  - `:encode_content_type` - content-type to be used in request header
+  - `:decode_content_types` - list of additional decodable content-types
+  - `:engine_opts` - optional engine options
+  """
+
+  @behaviour Tesla.Middleware
+
+  @content_type "application/msgpack"
```
andrewhr, line 40: Hey @chaodhib, maybe instead of reusing the same default attribute for encode/decode, what about having a list for decodable ones (similar to JSON middleware) with both `application/msgpack` and `application/x-msgpack`?

The rationale to support both out of box (sans overrides) is the first is what often we find on the wild but the [second arguably valid as well][1].

What do you think?

[1]: https://github.com/msgpack/msgpack/issues/194

chaodhib: Sure, I can add that

chaodhib: Done!

teamon:
🔥 🔥 🔥 💜 

@elixir-tesla/core Please take a look.

lib/tesla/middleware/message_pack.ex
```diff
@@ -0,0 +1,160 @@
+defmodule Tesla.Middleware.MessagePack do
+  @moduledoc """
```
yordis, line 2:
```suggestion
if Code.ensure_loaded?(Msgpax) do
  defmodule Tesla.Middleware.MessagePack do
    @moduledoc """
```

Maybe we should ensure that `Msgpax` is part of the project

chaodhib: done

yordis:
🚀 💜

PR #512, opened 2022-01-04
Update missing encodeJson middleware in Readme

epinault:
Looking at the code I found the middleware but was also confused why it existed. The decode was documented, not the encode. so clearing the confusion


teamon:
Thanks! 💜

PR #513, opened 2022-01-05
chore: move modules to its own file

yordis:
Just move the modules to different files, easier to find where the source code was for the given module.


lib/tesla/builder.ex
```diff
@@ -214,7 +214,7 @@ defmodule Tesla.Builder do
       for bang <- [:safe, :bang],
           client <- [:client, :noclient],
           opts <- [:opts, :noopts],
-          method in only && not (method in except) do
+          method in only && method not in except do
```
yordis, line 217: For some reason the formatter did this

yukster:
Yay! 🎉

teamon:
I'm not against this change, but it'll have to be coordinated to not cause too many conflicts. Let's reopen this open once we cleanup open PRs 💪

PR #514, opened 2022-01-10
Handle POST requests with `nil` body

### LINKED ISSUE #385: Same request fails with some adapters and succeeds with others
Hello,
I am migrating some of our requests from `HTTPoison` to `Tesla` and I found some requests are failing if I use `httpc` or `Mint` as the adapters, while `Hackney` works. I understand that it is because of HTTPoison using Hackney under the hood.

It would be nice, however to see some sort of uniformity in handling requests so that switching to a new adapter can be as seamless as possible. 

```elixir
uri = <URI>
method = :post
body = nil
headers: [
  {"Content-type", "application/json"},
  {"Accept", "application/json"},
  {"Authorization",
   "Bearer <TOKEN>"}
]
query = []

Tesla.request(client(),
   method: method,
   url: uri,
   body: body,
   headers: headers,
   query: query
)

defp client() do
    Tesla.client([Tesla.Middleware.JSON])
end
```

If I run that with Hackney it works well, if I use `httpc` for example I get this error

```elixir
** (FunctionClauseError) no function clause matching in :httpc.request/5

    The following arguments were given to :httpc.request/5:

        # 1
        :post

        # 2
        {'<URL>',
         [
           {'Content-type', 'application/json'},
           {'Accept', 'application/json'},
           {'Authorization',
            'Bearer <TOKEN>'}
         ]}

        # 3
        [autoredirect: false]

        # 4
        []

        # 5
        :default

    (inets) httpc.erl:149: :httpc.request/5
    (tesla) lib/tesla/adapter/httpc.ex:52: Tesla.Adapter.Httpc.request/2
    (tesla) lib/tesla/adapter/httpc.ex:22: Tesla.Adapter.Httpc.call/2
    (tesla) lib/tesla/middleware/json.ex:54: Tesla.Middleware.JSON.call/3
```

gaiabeatrice:
Should resolve #385.

This is my first time making a contribution. Let me know if there is anything I should do differently.

### DIFF
--- a/lib/tesla/adapter/gun.ex
+++ b/lib/tesla/adapter/gun.ex
@@ -180,7 +180,7 @@ if Code.ensure_loaded?(:gun) do
         Tesla.Adapter.Shared.format_method(env.method),
         Tesla.build_url(env.url, env.query),
         format_headers(env.headers),
-        env.body || "",
+        format_request_body(env),
         Tesla.Adapter.opts(
           [close_conn: true, body_as: :plain, send_body: :at_once, receive: true],
           env,
@@ -553,5 +553,10 @@ if Code.ensure_loaded?(:gun) do
           ip
       end
     end
+
+    # it is reasonable to make a post request with an empty body,
+    # however gun does not natively support it
+    defp format_request_body(%{method: :post} = env), do: env.body || "{}"
+    defp format_request_body(env), do: env.body || ""
   end
 end
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -54,7 +54,7 @@ defmodule Tesla.Adapter.Httpc do
         Tesla.build_url(env.url, env.query) |> to_charlist,
         Enum.map(env.headers, fn {k, v} -> {to_charlist(k), to_charlist(v)} end),
         content_type,
-        env.body,
+        format_request_body(env),
         Keyword.split(opts, @http_opts)
       )
     )
@@ -106,4 +106,9 @@ defmodule Tesla.Adapter.Httpc do
 
   defp handle({:error, {:failed_connect, _}}), do: {:error, :econnrefused}
   defp handle(response), do: response
+
+  # it is reasonable to make a post request with an empty body,
+  # however httpc does not natively support it
+  defp format_request_body(%{method: :post} = env), do: env.body || ""
+  defp format_request_body(env), do: env.body
 end
--- a/test/support/adapter_case/basic.ex
+++ b/test/support/adapter_case/basic.ex
@@ -96,6 +96,18 @@ defmodule Tesla.AdapterCase.Basic do
 
           assert {:error, _} = call(request)
         end
+
+        test "nil body in POST request" do
+          request = %Env{
+            method: :post,
+            url: "#{@http}/post",
+            body: nil,
+            headers: [{"content-type", "text/plain"}]
+          }
+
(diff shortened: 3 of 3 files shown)


lib/tesla/adapter/gun.ex
```diff
@@ -553,5 +553,10 @@ if Code.ensure_loaded?(:gun) do
           ip
       end
     end
+
+    # it is reasonable to make a post request with an empty body,
+    # however gun does not natively support it
+    defp format_request_body(%{method: :post} = env), do: env.body || "{}"
```
teamon, line 559: @alex-strizhakov What would be the correct solution here? Using `{}` doesn't seem 100% _right_

lib/tesla/adapter/httpc.ex
```diff
@@ -100,4 +100,9 @@ defmodule Tesla.Adapter.Httpc do
 
   defp handle({:error, {:failed_connect, _}}), do: {:error, :econnrefused}
   defp handle(response), do: response
+
+  # it is reasonable to make a post request with an empty body,
+  # however httpc does not natively support it
+  defp format_request_body(%{method: :post} = env), do: env.body || ""
```
teamon, line 106: Does it only apply to POST requests?

gaiabeatrice: Honestly I'm not sure about that.

When I first opened the issue, I ran into this problem with some Auth0 POST requests.
I can't really exclude other actions to have the same problem, so I was thinking about maybe starting here and see if this happens in other situations?
Or if you have other preferences I can explore these.

lib/tesla/adapter/gun.ex
alex-strizhakov: I don't get why `gun` doesn't support `nil` or `""` values for body requests.
```elixir
client = Tesla.client([Tesla.Middleware.JSON], Tesla.Adapter.Gun)
Tesla.post(client, "http://httpbin.org/post", "", headers: [{"content-type", "application/json"}])
```
```{:ok,
 %Tesla.Env{
   __client__: %Tesla.Client{
     adapter: {Tesla.Adapter.Gun, :call, [[]]},
     fun: nil,
     post: [],
     pre: [{Tesla.Middleware.JSON, :call, [[]]}]
   },
   __module__: Tesla,
   body: %{
     "args" => %{},
     "data" => "",
     "files" => %{},
     "form" => %{},
     "headers" => %{
       "Content-Length" => "0",
       "Content-Type" => "application/json",
       "Host" => "httpbin.org",
       "X-Amzn-Trace-Id" => "Root=1-6293805a-6d06c446640ad7a776aae5a7"
     },
     "json" => nil,
     "origin" => "176.232.61.138",
     "url" => "http://httpbin.org/post"
   },
   headers: [
     {"date", "Sun, 29 May 2022 14:16:58 GMT"},
     {"content-type", "application/json"},
     {"content-length", "332"},
     {"connection", "keep-alive"},
     {"server", "gunicorn/19.9.0"},
     {"access-control-allow-origin", "*"},
     {"access-control-allow-credentials", "true"}
   ],
   method: :post,
   opts: [],
   query: [],
   status: 200,
   url: "http://httpbin.org/post"
 }}
```
Same goes for `nil`.
@gaiabeatrice Can you please elaborate on what's wrong with `gun`?

yordis:
Hello 👋🏻 I am not sure where this is at, but @gaiabeatrice do you mind rebasing? 🙏🏻

PR #515, opened 2022-01-26
Update testing example

OnigiriJack:
I did not think it was clear to use `MyApp.get/1` since the function was not defined in a previous example.


teamon:
Good catch!

It should've been `MyApi` instead of `MyApp` as that's what's used in the config example above

yordis (approved):

PR #517, opened 2022-01-29
Update Tesla.Middleware documentation with functioning example code

yukster:
The current [example code](https://hexdocs.pm/tesla/Tesla.Middleware.html#module-examples) in the Tesla.Middleware documentation will not work as it is. The `inspect_headers` function does not return the `env` arg, which is what `Tesla.run` expects as the first arg. This results in this confusing error:

```** (ArgumentError) you attempted to apply a function named :opts on []. If you are using Kernel.apply/3, make sure the module is an atom. If you are using the dot syntax, such as map.field or module.function(), make sure the left side of the dot is an atom or a map```

In addition, the result of running the HTTP request will be an :ok/:error tuple, not the bare `Tesla.env` struct, so you need a different function (or different args) to handle that case.

Happy to tweak the approach here if desired but it seems like it would be ideal for example code to work correctly.

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -89,32 +89,19 @@ defmodule Tesla.Middleware do
 
   ### Examples
 
-      defmodule MyProject.InspectHeadersMiddleware do
-        @behaviour Tesla.Middleware
+    defmodule MyProject.InspectHeadersMiddleware do
+      @behaviour Tesla.Middleware
 
-        @impl Tesla.Middleware
-        def call(env, next, options) do
-          env
-          |> pre_inspect_headers(options)
-          |> Tesla.run(next)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by teamon** [comment]
Good catch!

The _idiomatic_ middleware would be something like this

```ex
defmodule MyProject.InspectHeadersMiddleware do
  @behaviour Tesla.Middleware
   
  @impl true
  def call(env, next, options) do
     IO.inspect(env.headers, options)
    
    with {:ok, env} <- Tesla.run(env, next) do
      IO.inspect(env.headers, options)
      {:ok, env}
    end
  end
end
```

yukster:
@teamon updated with your code. Thanks!

yordis:
🚀

PR #518, opened 2022-01-30
Update README.md

idkjs:
I tried running the example. Looks like the target url has changed. Fixed example code so it runs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ defmodule GitHub do
   plug Tesla.Middleware.JSON
 
   def user_repos(login) do
-    get("/users/" <> login <> "/repos")
+    get(login <> "/repos")
   end
 end
 ```


README.md
```diff
@@ -32,7 +32,7 @@ defmodule GitHub do
   plug Tesla.Middleware.JSON
 
   def user_repos(login) do
-    get("/users/" <> login <> "/repos")
+    get(login <> "/repos")
```
entertainyou, line 35: The original version seems correct, 
`https://api.github.com/users/teamon/repos` returns expected response while the one without `users` `https://api.github.com/teamon/repos` returns 404

teamon:
GitHub now requires `user-agent` header and authorization. Do you know any other API that could serve as the basic example?

README.md
yordis: This will break, the endpoint is `https://api.github.com/users/{user handler}/repos`

yordis:
The existing change will break the example since the URL is wrong.

I am gonna close it, for now, my apologies if it was the wrong call, feel free to reopen it.

PR #520, opened 2022-02-27
Allow set custom formatter for logger

oo6:
Sometimes we need a custom formatter to log the request body or something similar.

Heavily based on Elixir's Logger.Formatter


teamon:
Awesome! 💜

PR #523, opened 2022-03-13
Add Client.update_middleware/2 function

maciej-szlosarczyk:
I have seen this being a pain point a few times so maybe it's useful for someone else.

A typical use-case is the following:

A third-party library (i.e GitHub) exposes a function that returns a Tesla client, but I want to append a header or add logging/telemetry to it. Currently I need to write a wrapper for Tesla.Client.middleware/1 and Tesla.client/2, this merges the two into a single function.

### DIFF
--- a/lib/tesla/client.ex
+++ b/lib/tesla/client.ex
@@ -43,6 +43,25 @@ defmodule Tesla.Client do
     unruntime(client.pre)
   end
 
+  @doc ~S"""
+  Given an update function, create a new client by transforming the
+  list of middlewares for an existing one.
+
+  ## Examples
+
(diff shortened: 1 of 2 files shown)


teamon:
👋 

I've found similar need when working on the OpenAPI generator and I've ended up with https://github.com/elixir-tesla/tesla/blob/openapi/lib/tesla/client.ex#L51-L72 (this code lives only in the `openapi` branch, nowhere near to be released yet)

Do you see the need for generic `update_middleware` or would `insert` be enough?

maciej-szlosarczyk:
Hey @teamon, for my use-case insert would work as well. Let me know if I should extract it from that branch and rework this PR with that or help in any other way.

PR #526, opened 2022-04-30
Gun library version updated

kingdomcoding:


teamon:
Hi!

In general, libraries should not depend on strict version of other libraries, especially when the version is only RC (Release Candidate).

I think we can try to relax the dependency to something like: `">= 1.3"` and check if everything works well with gun 2.x

kingdomcoding:
Using a `>= 1.3.3` does not resolve the problem.

This still installs the 1.3.3 version.

The [telegram library](https://github.com/visciang/telegram) depends on tesla but won't successfully resolve dependencies unless `gun` is at least at 2.0.

Does this mean we need to wait till `gun` releases a stable 2.0 version?

alex-strizhakov:
@kingdomcoding you can use `override` flag and `ref` for needed commit to use gun from GitHub directly (at your's own risk)
```elixir
{:gun, github: "ninenines/gun", ref: "f9175998687678e227bdd49669e2d83f0648fa57", override: true}
```

I'm using `gun` from master in several projects and it's stable :)

fuelen:
gun 2.0 has been released

PR #528, opened 2022-05-06
Uses `:max_body` in `hackney` Adapter

thiagopromano:


teamon:
@thiagopromano This looks good! Could you add a simple test case & add the new option to Hackney adapter moduledoc?

yordis (approved):

PR #534, opened 2022-06-17
Filter sensitive data from inspect(Tesla.Client.t)

maltoe:
Hi :wave: 

We recently discovered - luckily before going to production - that we leaked secret tokens for an API integration, as we were doing something like this

```elixir
{:ok, _} = 
  [{Tesla.Middlewares.BearerAuth, token: "supersecret"}]
  |> Tesla.client()
  |> Tesla.post(...)
```

The resulting `MatchError` in case of network failures etc. would include the inspected `Tesla.Client` struct which in turn included the `supersecret` token.

We thought implementing `Inspect` within Tesla itself and filtering some options would be a more sensible default for all users, hence this PR.

Best,
malte

## Todos

- [ ] Add `Tesla.Middleware.BasicAuth.Options` struct
- [ ] Add `Tesla.Middleware.BearerAuth.Options` struct
- [ ] Add `Tesla.Middleware.DigestAuth.Options` struct

### DIFF
--- a/test/tesla/client_test.exs
+++ b/test/tesla/client_test.exs
@@ -61,7 +61,9 @@ defmodule Tesla.ClientTest do
       inspected = middlewares |> Tesla.client() |> inspect()
 
       refute String.contains?(inspected, "secret")
-      assert String.contains?(inspected, "OK")
+      assert String.contains?(inspected, ~s(password: "[FILTERED]"))
+      assert String.contains?(inspected, ~s(username: "[FILTERED]"))
+      assert String.contains?(inspected, ~s(token: "[FILTERED]"))
     end
   end
 end
--- a/test/tesla/client_test.exs
+++ b/test/tesla/client_test.exs
@@ -49,4 +49,19 @@ defmodule Tesla.ClientTest do
       assert middlewares == Tesla.Client.middleware(client)
     end
   end
+
+  describe "Inspect.Tesla.Client" do
+    test "ensures that no secrets are leaked in logs" do
+      middlewares = [
+        {Tesla.Middleware.BasicAuth, username: "secret", password: "secret", other: "OK"},
+        {Tesla.Middleware.BearerAuth, token: "secret", other: "OK"},
+        {Tesla.Middleware.DigestAuth, %{username: "secret", password: "secret", other: "OK"}}
+      ]
+
+      inspected = middlewares |> Tesla.client() |> inspect()
+
+      refute String.contains?(inspected, "secret")
+      assert String.contains?(inspected, "OK")
+    end
+  end
 end
--- a/lib/tesla/client.ex
+++ b/lib/tesla/client.ex
@@ -48,44 +48,45 @@ defmodule Tesla.Client do
   defp unruntime({module, :call, [opts]}) when is_atom(module), do: {module, opts}
   defp unruntime({:fn, fun}) when is_function(fun), do: fun
 
-  defimpl Inspect do
-    @filtered "[FILTERED]"
-
-    @sensitive_opts %{
-      Tesla.Middleware.BasicAuth => [:username, :password],
-      Tesla.Middleware.BearerAuth => [:token],
-      Tesla.Middleware.DigestAuth => [:username, :password]
-    }
-
-    def inspect(%Tesla.Client{} = client, opts) do
-      client
-      |> Map.update!(:pre, &filter_sensitive_opts/1)
-      |> Inspect.Any.inspect(opts)
-    end
-
-    defp filter_sensitive_opts(middlewares) do
-      Enum.map(middlewares, fn
-        {middleware, :call, [opts]} ->
-          sensitive_opts = Map.get(@sensitive_opts, middleware, [])
-          filtered_opts = Enum.reduce(sensitive_opts, opts, &maybe_redact(&2, &1))
-          {middleware, :call, [filtered_opts]}
-
-        middleware ->
-          middleware
-      end)
-    end
-
-    defp maybe_redact(opts, key) do
-      cond do
-        is_map(opts) and Map.has_key?(opts, key) ->
-          Map.put(opts, key, @filtered)
-
-        is_list(opts) and Keyword.has_key?(opts, key) ->
-          Keyword.put(opts, key, @filtered)
-
-        true ->
-          opts
-      end
-    end
-  end
+  # TODO: Remove once we are done with all the middlewares
+  #  defimpl Inspect do
+  #    @filtered "[FILTERED]"
+  #
+  #    @sensitive_opts %{
+  #      Tesla.Middleware.BasicAuth => [:username, :password],
+  #      Tesla.Middleware.BearerAuth => [:token],
+  #      Tesla.Middleware.DigestAuth => [:username, :password]
+  #    }
+  #
+  #    def inspect(%Tesla.Client{} = client, opts) do
+  #      client
+  #      |> Map.update!(:pre, &filter_sensitive_opts/1)
+  #      |> Inspect.Any.inspect(opts)
+  #    end
+  #
+  #    defp filter_sensitive_opts(middlewares) do
+  #      Enum.map(middlewares, fn
+  #        {middleware, :call, [opts]} ->
+  #          sensitive_opts = Map.get(@sensitive_opts, middleware, [])
+  #          filtered_opts = Enum.reduce(sensitive_opts, opts, &maybe_redact(&2, &1))
+  #          {middleware, :call, [filtered_opts]}
+  #
+  #        middleware ->
+  #          middleware
+  #      end)
+  #    end
+  #
+  #    defp maybe_redact(opts, key) do
+  #      cond do
+  #        is_map(opts) and Map.has_key?(opts, key) ->
+  #          Map.put(opts, key, @filtered)
+  #
+  #        is_list(opts) and Keyword.has_key?(opts, key) ->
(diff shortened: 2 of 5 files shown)


yordis:
> that we leaked secret tokens for an API integration

Do you mind expanding on that, do you mean you leaked to some 3rd party provider?

---- 

I am not opposed to the idea.

The curious part is that this issue would be all over Elixir code bases, so I am not sure what is the right call here.

As far as I can tell you would modify your logger to hide sensitive information, but I may be wrong.

Gonna lean on @teamon for some feedback about it.

test/tesla/client_test.exs
```diff
@@ -49,4 +49,19 @@ defmodule Tesla.ClientTest do
...
+      middlewares = [
+        {Tesla.Middleware.BasicAuth, username: "secret", password: "secret", other: "OK"},
+        {Tesla.Middleware.BearerAuth, token: "secret", other: "OK"},
+        {Tesla.Middleware.DigestAuth, %{username: "secret", password: "secret", other: "OK"}}
+      ]
+
+      inspected = middlewares |> Tesla.client() |> inspect()
+
+      refute String.contains?(inspected, "secret")
+      assert String.contains?(inspected, "OK")
```
kyleVsteger, line 64: I'd probably add an assertion to ensure `[FILTERED]` appears twice

teamon:
I tend to have a custom implementation of Inspect protocol on a per-project basis, often something as simple as:

```ex
defimpl Inspect, for: Tesla.Client do
  @moduledoc """
  Implement custom Inspect protocol for Tesla.Client
  to prevent leaking secrets added with Tesla.Middleware.Headers
  to AppSignal or logs.
  """

  import Inspect.Algebra

  def inspect(client, opts) do
    attrs = client |> Map.take([:adapter]) |> Map.to_list()
    concat(["#Tesla.Client<", to_doc(attrs, opts), ">"])
  end
end
```

The problem with implementing it directly into tesla package is that as far as I know one wouldn't be able to override it anymore.

test/tesla/client_test.exs
yordis: @kyleVsteger since he refuted `secret` and I modified a bit the test case, we can avoid checking how many of them appear

yordis:
A solid argument is that Ecto does the same https://github.com/elixir-ecto/ecto/blob/d52039037ae8d3ba66034c35c08fc4d56c621447/lib/ecto/changeset.ex#L3134-L3172 (from Kyle Steger)

kyleVsteger:
> A solid argument is that Ecto does the same https://github.com/elixir-ecto/ecto/blob/d52039037ae8d3ba66034c35c08fc4d56c621447/lib/ecto/changeset.ex#L3134-L3172 (from Kyle Steger)
> 
> but then as @teamon pointed out, what happens when we have other middleware that wants to accomplish the same? We can't redact anymore something?

It's possible to write another middleware, `Tesla.RedactValues` or something of the like that takes a list of keys to redact. It looks like `Tesla.post|patch|etc` accept `[option]`, which may be an OK place to add `redact_keys: [:a, :b, :c]`. 

I think what is proposed here provides immediate benefit and prevents known secret data from being leaked in logs. Gets a thumb in my book

maltoe:
@yordis 

> Do you mind expanding on that, do you mean you leaked to some 3rd party provider?

Leaked to our application monitoring as part of its exception handling.

> it directly into tesla package is that as far as I know one wouldn't be able to override it anymore

How about hiding it behind a compile-time flag? I'd assume that the vast majority of users do not have a custom implementation, so defaulting it to `true` would make sense, IMO. Safety first.

```
config :tesla, implement_inspect_protocol: false
```

maltoe:
Also, as a side note: We have since discovered that we also use the `KeepRequest` middleware in one case, so we would also need to purge the `req_headers` from the `opts` field of `Tesla.Env` :/ - We're currently debating whether we want to implement `Inspect` for `Tesla.Env` as well (and optionally upstream it, too), or whether we can get rid of `KeepRequest` in this case. Implementation of `Inspect` for `Tesla.Env` feels a bit brittle in this case. What do others think?

maltoe:
@yordis 

> As far as I can tell you would modify your logger to hide sensitive information, but I may be wrong.

That's unfortunately not possible in this case. Exception is raised, Appsignal `inspect()`s it and we can't really filter it anymore from the resulting string.

yordis:
>  Safety first.

Agree with this, but having a solution when opting out without so much rework should be taken into consideration as well.

> What do others think?

What if the input is an struct instead of a keyword list, then you implement the redacting for such struct, example:

```elixir
{MyMiddlewareName, %MyMiddlewareName.Config{token: "..."}}
```

So you implement the protocol for `MyMiddlewareName.Config`

Would that work?

If that strategy work, we could decide to commit to what we have, and maybe recommend people to do such a thing if they want to control redacting things.

Thoughts?

maltoe:
Gentlemen, any update on this? I'm happy to rework it in any way you conceive, but I do believe the status quo is an unnecessary pitfall / security risk for unaware people.

yordis:
@maltoe sorry it took me this long to get back to this issue, but I have other duties.

From my perspective, the tricky situation with this PR is that we are taking over control of the `Inspect` for the entire `Tesla.Client.`

From the library perspective,

How do we ensure that this implementation of the `Inspect` protocol will work for all middleware (including the ones we did not write)?

That may not be a concern, but I am unsure.

This leads us to two possible solutions,

1. In your application code, you define the `Inspect` protocol (less ideal)
2. Use a struct as configuration (see my changes)


## IMPORTANT

I pushed some changes showcasing the usage of `Tesla.Middleware.BasicAuth.Options`, which I made the rookie mistake of force-push because I rebased (my bad).

If you wish to continue helping in the subject, we must do the same for `Tesla.Middleware.BearerAuth` and `Tesla.Middleware.DigestAuth` thus far.

Let me know if you can help with the task. Otherwise, I can take over.

maltoe:
@yordis 

> sorry it took me this long to get back to this issue, but I have other duties.

Sorry if my last comment came across too pushy. I'm sure you have and thank you for your work on tesla, it's an awesome tool :hammer_and_wrench: :purple_heart:  !

> From my perspective, the tricky situation with this PR is that we are taking over control of the Inspect for the entire Tesla.Client.
> From the library perspective, how do we ensure that this implementation of the Inspect protocol will work for all middleware (including the ones we did not write)?

Personal opinion: I'm wondering whether we're "overthinking" it - the original code in this PR was my attempt at fixing the "leak" while leaving some useful information in the `inspect` output. We have since refined it in our codebase and even expanded it to `Tesla.Env` (`req_headers` in case `KeepRequest` is used). But I'm more sceptical about its inclusion in Tesla itself now, as it has become a non-negligible amount of code, still doesn't cover all cases (e.g. third-party middlewares as you mentioned), and especially as I can't remember ever having needed the information in the `inspect` output at all.

Hence, how about going back to what @teamon originally said in https://github.com/elixir-tesla/tesla/pull/534#issuecomment-1164551535 and providing an extremely basic default implementation (e.g. hiding all middlewares including the internal ones) with an opt-out? That way you prevent leaks even from external middlewares, protect the innocents, and people don't have to rework much at all when they opt-out :grimacing: 

Otherwise, I also like your struct approach (though `req_headers` in `Tesla.Env` needs to be considered as well). Unfortunately, I don't know how much time I'll be given to work on this any time soon, so can't promise much.

Thanks again for considering & best wishes.

yordis:
Coming back to this,

We should refrain from doing anything about this, which is a fundamental problem for Elixir. In the best case, a scapegoat exists, and changing the behavior of the Inspect is good enough; adding `[redacted]` is the tricky situation here, as anything in the stack could be printed out.

This is also why Logger filters exist: You can clean up and remove things there to ensure that you do not leak information.

I am not opposed 100% yet to trying to fix it, but looking at the code and whatnot, I am somewhat thinking of pushing the responsibility to copy+paste @teamon example or simply use the existing tool from Elixir, such as the Logger filters to deal with the situation

teamon:
I've recently ended up with something like this:

```ex
defimpl Inspect, for: Tesla.Client do
  @moduledoc """
  Implement custom Inspect protocol for Tesla.Client
  to prevent leaking secrets added to headers
  """

  @conceal [
    {Tesla.Middleware.BasicAuth, [:password]},
    {Tesla.Middleware.BearerAuth, [:token]},
    {Tesla.Middleware.DigestAuth, [:password]}
  ]

  def inspect(client, opts) do
    client =
      Map.update!(client, :pre, fn middlewares ->
        Enum.map(middlewares, fn {module, fun, [opts]} ->
          case @conceal[module] do
            nil ->
              {module, fun, opts}

            attrs ->
              {module, fun,
               Enum.reduce(attrs, opts, fn attr, acc ->
                 acc
                 |> Access.get_and_update(attr, fn val -> {val, "*****"} end)
                 |> elem(1)
               end)}
          end
        end)
      end)

    Inspect.Map.inspect(client, opts)
  end
end
```

Maybe there is a room for some `Tesla.conceal` function that could be used as the implementation if one want's to change the Inspect behaviour? (which is not the only way to do it as discussed and should not be changed by default)

yordis (commented):
The problem of leaking secrets in logs is a well-known issue throughout the Elixir ecosystem, not just with Tesla.

After many years of observing other packages and considering different approaches to addressing this issue, I have decided not to accept the pull request (for now). Instead, it's advisable to add filters to your Logger to remove any unwanted data.

Maybe in v2, we come up with a different API that solves the issue.

Thank you so much for your contribution, and my apologies for taking too long.

PR #539, opened 2022-08-17
chore: upgrade deps to latest versions

yordis:


epinault:
I know ibrowse is blocking, any change we can upgrade it separately and let everything else go through?

yordis:
>  Why not using Dependabot?

Yeah, that is the intent long-term!

PR #550, opened 2022-11-29
fix httpc bodyless methods

olafura:
Currently we are creating a request with body and content_type in many cases for methods that can't support it. Httpc doesn't know what to do an outputs a function match error.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -18,6 +18,4 @@ if Mix.env() == :test do
     sasl_error_logger: false
 
   config :tesla, MockClient, adapter: Tesla.Mock
-
-  config :tesla, Tesla.Adapter.Mint, cacert: ["./deps/httparrot/priv/ssl/server-ca.crt"]
 end
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -69,6 +69,12 @@ defmodule Tesla.Adapter.Httpc do
     :httpc.request(method, {url, headers}, http_opts, opts)
   end
 
+  # These methods aren't able to contain a content_type and body
+  defp request(method, url, headers, _content_type, _body, {http_opts, opts})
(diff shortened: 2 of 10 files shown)


yordis:
I remember some discussions about it, and I remember something around the point of "HTTP" spec technically allowing a body.

@teamon I am gonna need your help in terms of what we should do here, I feel we should merge it, but still, different clients do different things so a bit tricky.

- https://stackoverflow.com/questions/978061/http-get-with-request-body
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET

olafura:
@yordis didn't know before this that GET could contain a body and I used to use telnet to send request for fun.

The reason why I implemented this is because httpc can't contain a body for anything but post, put, patch, and delete.
https://github.com/erlang/otp/blob/40922798411c2d23ee8a99456f96d6637c62b762/lib/inets/src/http_client/httpc.erl#L161

Reading more into the links provided where they mention that servers can reject the request sometimes with the body. It's an edge case but if we really do want to support it we will have to patch httpc to be able to support it.

Reading through the source code of hackney they do seem to support it for some times of body requests but not others.

yordis:
I know you are 100% right on this one, httpc would need to fix their stuff.

So I guess people must be aware of such technicality, otherwise 🤷🏻

yordis:
CI keeps failing 😢

olafura:
Yeah don't know why the tests are trying to access priv directly and not through `:code.priv_dir`

olafura:
I sound really judgemental :grin:, I came to Elixir through Erlang ( was still getting tripped up by it's syntax ). So I've been exposed to a lot of the old best practices.

yordis:
😭

olafura:
There is at least one flaky test

yordis:
I remember something about a flaky test for sure!

yordis (approved):

PR #557, opened 2023-02-02
Correctly indent code example

mhanberg:
The example was not being rendered correctly in the docs because it wasn't indented correctly.


yordis:
🚀 💜

PR #558, opened 2023-02-03
feat: allow to specify encoding strategy for query params

RudolfMan:
I was adapting Tesla in a project when we encode query params with `:rfc3986` and I noticed that, while Tesla provides conveniences to build url encoding query params for us, it doesn't allow to specify the encoding strategy.

### TL;DR

This PR allows specifying option `query_encoding: :rfc3986` so that whitespaces in query params will be encoded as "%20".

```elixir
Tesla.get("http://example.com", query: [username: "John Smith"], opts: [query_encoding: :rfc3986])
```
will build the following url
```
http://example.com?username=John%20Smith
```

Also adds optional argument to `Tesla.build_url` and `Tesla.encode_query`.

----
Tesla uses [URI.encode_query/2](https://hexdocs.pm/elixir/1.14.3/URI.html#encode_query/2) to encode query params. Since Elixir 1.12 that function allows to specify the encoding strategy.

> You can specify one of the following encoding strategies:
>
>  - `:www_form` - (default, since v1.12.0) keys and values are URL encoded as per [encode_www_form/1](https://hexdocs.pm/elixir/1.14.3/URI.html#encode_www_form/1). This is the format typically used by browsers on query strings and form data. It encodes " " as "+".
>
>  - `:rfc3986` - (since v1.12.0) the same as `:www_form` except it encodes " " as "%20" according [RFC 3986](https://tools.ietf.org/html/rfc3986). This is the best option if you are encoding in a non-browser situation, since encoding spaces as "+" can be ambiguous to URI parsers. This can inadvertently lead to spaces being interpreted as literal plus signs.
>
> Encoding defaults to `:www_form` for backward compatibility.

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -346,7 +346,7 @@ defmodule Tesla do
   @spec build_url(Tesla.Env.t()) :: String.t()
   def build_url(%Tesla.Env{} = env) do
     query_encoding = Keyword.get(env.opts, :query_encoding, :www_form)
-    Tesla.build_url(url, query, encoding)
+    Tesla.build_url(env.url, env.query, encoding)
   end
 
   def encode_query(query, encoding \\ :www_form) do
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -296,6 +296,10 @@ defmodule Tesla do
 
   Useful when you need to create an URL with dynamic query params from a Keyword list
 
+  Allows to specify the `encoding` strategy to be one either `:www_form` or `:rfc3986`
+  Read more about options in [`URI.encode_query/2`](https://hexdocs.pm/elixir/1.14.3/URI.html#encode_query/2)
+  Defaults to `:www_form`, ignored when compiled with elixir version older than 1.12.
+
   ## Examples
 
       iex> Tesla.build_url("http://api.example.com", [user: 3, page: 2])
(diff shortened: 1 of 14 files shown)


yordis:
Related to: https://github.com/elixir-tesla/tesla/issues/538

lib/tesla.ex
```diff
@@ -517,19 +521,34 @@ defmodule Tesla do
...
+      iex> Tesla.build_url("http://api.example.com", [user_name: "John Smith"])
+      "http://api.example.com?user_name=John+Smith"
+
+      # specified encoding strategy :rfc3986
+      iex> Tesla.build_url("http://api.example.com", [user_name: "John Smith"], :rfc3986)
+      "http://api.example.com?user_name=John%20Smith"
   """
-  @spec build_url(Tesla.Env.url(), Tesla.Env.query()) :: binary
-  def build_url(url, []), do: url
+  @spec build_url(Tesla.Env.url(), Tesla.Env.query(), :rfc3986 | :www_form) :: binary
```
yordis, line 532:
```suggestion
  @type encoding_strategy :: :rfc3986 | :www_form

  @spec build_url(Tesla.Env.url(), Tesla.Env.query(), encoding_strategy) :: binary
```

Also related to this, what happen if some weird encoding is needed? 

As an example: https://github.com/elixir-tesla/tesla/issues/538#issuecomment-1242348413 I had to do that for a project, and, I am not too familiar with whatever RFCs exists out there.

So, what I created was part of some RFC?
Should we allow some callback to handle the encoding as a scape goat? Maybe add it when somebody asks for it? What would be the workload then?

yordis (changes requested):

yordis:
Thank you so much for taking the time for this. I appreciate it.

yordis:
Any final thoughts related to our conversation?

yordis (approved):

yordis:
@teamon share some love over here! Is the function changing from arity 2 to arity 3 with the default value any concern here?

teamon (approved):
This is all backwards compatible, looks good to me!

PR #560, opened 2023-02-22
Adds docs to the Mint adapter on 'transport_opts'

JesseHerrick:
While looking into an API that required using special options passed into `transport_opts`, I noticed that Mint supports it but that Tesla had no docs on whether `transport_opts` was a valid option for this adapter. Looking at the code, Tesla does support passing `transport_opts` into Mint. I think this would be a useful addition to the docs.


yordis:
🚀 💜

teamon:
Thank you! 💜💜💜

PR #568, opened 2023-03-31
OTel - Add context propagation

### LINKED ISSUE #567: Tesla.Middleware.Timeout and context propagation
## Context

In order to be able to propagate the OpenTelemetry context across children processes, we need to inject the parent `traceparent` into the child. Timeout middleware is spawning a new child [here](https://github.com/elixir-tesla/tesla/blob/v1.5.1/lib/tesla/middleware/timeout.ex#L43) and not propagating the context. There is already some wrappers for doing so, as defined [here](https://hexdocs.pm/opentelemetry_process_propagator/OpentelemetryProcessPropagator.Task.Supervisor.html).

## Proposed solution

I open this issue to discuss what would be better for this middleware:
- Add an optional dependency to the `opentelemetry_process_propagator` and use the wrapper. We can basically check if the propagator exists and use the wrapper, otherwise use the current impl.
- Implement the wrapper in this middleware , but it will require also some dependencies on how the injection is done.

I think the first would be better as we take benefit of the changes from the other library. If it makes sense, I can open a PR with the changes.

I created a possible implementation [here](https://github.com/elixir-tesla/tesla/pull/568/files). cc/ @teamon 😁

carrascoacd:
An implementation example for https://github.com/elixir-tesla/tesla/issues/567

Adds a new middleware in order to make explicit that OTel will be used when plugged.

closes #567


lib/tesla/middleware/timeout.ex
```diff
@@ -21,6 +21,11 @@ defmodule Tesla.Middleware.Timeout do
 
   @default_timeout 1_000
 
+  # Optional. Used for context propagation from parent process to child process
+  @task_module if Code.ensure_loaded?(:opentelemetry_process_propagator),
+                 do: OpentelemetryProcessPropagator.Task,
+                 else: Task
+
```
yordis, line 28: Please move this as a configuration instead of assuming that `:opentelemetry_process_propagator` should be used.

yordis (changes requested):

lib/tesla/middleware/timeout.ex
carrascoacd: Thanks for your comment!

I'm not sure if sharing such a detail to the clients is necessary. In case we change the implementation we are coupling here. What are you trying to achieve with the suggestion? So I can think about an alternative. 

Another alternative could be writing a new middleware with explicit support. In that case I think this won't be the best place but the erlang contrib repository for OTel.

yordis: > I'm not sure if sharing such a detail to the clients is necessary.

It is necessary to expose the detail to the clients! Only some people want to use the propagator even when using OTEL, primarily when you communicate with systems that **you do not trust**, so you shouldn't be sending them internal information. **It is a security concern**.

> What are you trying to achieve with the suggestion?

Opt-in configuration:

```elixir
 defmodule MyClient do
    use Tesla
    plug Tesla.Middleware.Timeout,
      timeout: 2_000,
      task_module: OpentelemetryProcessPropagator.Task
  end
```

> Another alternative could be writing a new middleware with explicit support. In that case I think this won't be the best place but the erlang contrib repository for OTel.

You could, but such contrib packages exist primarily because the original packages want to avoid coupling to OTEL. I think adding the abstraction you proposed solves both issues. Maybe add some documentation so a simple copy+paste does the trick.

carrascoacd: > It is a security concern

You are totally right, I only had in mind the usage of the middleware in internal networks.

I think I'll push a copy of this in the contrib repository. Thanks!

yordis: > I think I'll push a copy of this in the contrib repository. Thanks!

As I said, keep here, you don't need to push to contrib. What I am trying to say is that it is a bad idea to be the default behavior

carrascoacd: @yordis please, take a look again. I've duplicated the middleware making it explicit OTel will be used if plugged.

yordis: Is there any reason why it can not be a configuration option?

```elixir
 defmodule MyClient do
    use Tesla
    plug Tesla.Middleware.Timeout,
      timeout: 2_000,
      task_module: OpentelemetryProcessPropagator.Task # this way!
  end
```

carrascoacd: Mainly because of some reasons:
- Injecting a `Task` module that has no common interface (`behaviour`) is not right. It is a bit tricky, you are assuming clients know what a task module is which is not a clear concept here.
- At the same time you are violating the open-closed principle, since if you want to change the implementation and use a `Task.Supervisor` you are coupling your clients here, not opening your code for changes.

So, if you are trying to DRY, I think that if we want to opt-in the "propagate context" the way of doing it should be different (focused on the feature, not in the implementation) i.e.:

```elixir
 defmodule MyClient do
    use Tesla
    plug Tesla.Middleware.Timeout,
      timeout: 2_000,
      otel_enabled: true # default false, by security concerns. You are hidding the Task detail here
  end
```

Does it make sense? I can make opt-in if you find it better, I have no strong oppinion here but exposing the concept of "task module".

yordis: Hey, @carrascoacd thank you for sharing more context about the topic. I appreciate your time.

I think you have good solid (you see what I did there?! 😄) points, but I think the alternative's simplicity is much better. I choose the latter to maintain two middleware that is identical vs. a straightforward and simple configuration.

That does not mean I am "right" and you are "wrong," just different perspectives. Hopefully, I am okay here. Time will tell.

yordis (approved):

lib/tesla/middleware/timeout.ex
carrascoacd: > That does not mean I am "right" and you are "wrong," just different perspectives. Hopefully, I am okay here. Time will tell.

I totally agree. I appreciate your comments, thanks for reviewing!

teamon:
Hey @carrascoacd 👋 

Thank you both for this 💜💜💜

PR #574, opened 2023-04-25
Fix mocked calls failing when using Timeout middleware

Cantido:
Since the Timeout middleware is creating a new process, the mock in the pdict is lost. Fortunately it's easy to copy the pdict over to the new process.

### DIFF
--- a/lib/tesla/middleware/timeout.ex
+++ b/lib/tesla/middleware/timeout.ex
@@ -40,7 +40,11 @@ defmodule Tesla.Middleware.Timeout do
     timeout = Keyword.get(opts, :timeout, @default_timeout)
     task_module = Keyword.get(opts, :task_module, Task)
 
-    task = safe_async(task_module, fn -> Tesla.run(env, next) end)
+    mock = Tesla.Mock.pdict_get()
+    task = safe_async(task_module, fn ->
+      if mock, do: Tesla.Mock.pdict_set(mock)
+      Tesla.run(env, next)
+    end)
 
     try do
       task
--- a/lib/tesla/mock.ex
+++ b/lib/tesla/mock.ex
@@ -225,8 +225,11 @@ defmodule Tesla.Mock do
     end
   end
 
-  defp pdict_set(fun), do: Process.put(__MODULE__, fun)
-  defp pdict_get, do: Process.get(__MODULE__)
+  @doc false
+  def pdict_set(fun), do: Process.put(__MODULE__, fun)
+
+  @doc false
+  def pdict_get, do: Process.get(__MODULE__)
 
   defp agent_set(fun) do
     case Process.whereis(__MODULE__) do
--- a/test/tesla/middleware/timeout_test.exs
+++ b/test/tesla/middleware/timeout_test.exs
@@ -31,6 +31,14 @@ defmodule Tesla.Middleware.TimeoutTest do
     end
   end
 
+  defmodule MockClient do
+    use Tesla
+
+    plug Tesla.Middleware.Timeout, timeout: 100
+
+    adapter Tesla.Mock
+  end
+
   defmodule DefaultTimeoutClient do
     use Tesla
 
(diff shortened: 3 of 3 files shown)


lib/tesla/middleware/timeout.ex
```diff
@@ -40,7 +40,11 @@ defmodule Tesla.Middleware.Timeout do
     timeout = Keyword.get(opts, :timeout, @default_timeout)
     task_module = Keyword.get(opts, :task_module, Task)
 
-    task = safe_async(task_module, fn -> Tesla.run(env, next) end)
+    mock = Tesla.Mock.pdict_get()
+    task = safe_async(task_module, fn ->
+      if mock, do: Tesla.Mock.pdict_set(mock)
+      Tesla.run(env, next)
+    end)
```
yordis, line 47: Hey there, this seems a no-go from my perspective, `Tesla.Mock` should `not` be used in production code as such.

We need to find a better alternative. In this particular case, you can swap the `Task` module implementation using `task_module` configuration.

yordis (changes requested):

lib/tesla/middleware/timeout.ex
Cantido: I can understand that, for sure. Unfortunately I'm using the OpenTelemetry task propagator too so this is the only thing that works for my use case; it's the only simple solution that works generically for all task modules. I know that it feels icky to use `Tesla.Mock` in a production module, but it's simple, it solves the problem, and it doesn't have any nasty side effects.

Maybe we could find a generic way to transfer state into the task that the production code can implement and that `Telsa.Mock` can use? Perhaps there's some way to build a feature into the adapter behavior, since maintaining state between processes is a requirement of `Tesla.Mock`?

yordis: We introduced a way to change the `Task Module` here https://github.com/elixir-tesla/tesla/pull/568

Maybe swap the implementation of the task module as following (kind of I hope you get the idea):

```elixir
# somewhere in the test/support dir
     defmodule MyClient do
        use Tesla
        plug Tesla.Middleware.Timeout,
          # notice here
          task_module: TestSupport.MockTaskModule
      end
```

Where the `TestSupport.MockTaskModule` could be something around the lines of:

```elixir
# inside test/support dir
defmodule TestSupport.MockTaskModule do
  def async(fn) do
     mock = Tesla.Mock.pdict_get()
     Task.async(fn ->
       # do it here
       Tesla.Mock.pdict_set(mock)
       {:ok, fn.()}
     end)
  end

  defdelegate await(task, timeout), to: Task
  defdelegate shutdown(task, timeout), to: Task
end

---

Also, what issue is this fixing per se? We dont have any failing tests, do we? Not sure what is broken, if you dont mind making it fail to see the issue.

Cantido: I added a test in `test/tesla/middleware/timeout_test.exs` that triggers this. The issue is that if you've set up a client with the `Timeout` middleware, mocks made in tests do not work, since the Timeout middleware moves everything to another process, and the `Tesla.Mock` adapter only checks the pdict for its current process.

I don't think I understand this solution; the Timeout middleware is in production code and I can't swap out the configured `task_module` module during tests, can I?

yordis: >  I can't swap out the configured task_module module during tests, can I?

**In the testing** environment, yeah, you can. Notice here https://github.com/elixir-tesla/tesla/blob/04a94e6a8ea04740833f0980336ea5e5d138cc0f/test/tesla/middleware/timeout_test.exs#L52-L70

Cantido: Are you suggesting that I use an entirely different client module during tests?

yordis: Yes, one that does the mock thingy you need in order to make the test to pass

PR #581, opened 2023-05-12
Remove "import Bitwise" to fix a warning

whatyouhide:


yordis:
Allo 👋🏻 should we bump https://github.com/elixir-tesla/tesla/blob/30c3a298bc0de841d8929055429598342ef75573/mix.exs#LL13C7-L13C24 then?

I was about to do it at some point to deprecate the support for older versions

whatyouhide:
@yordis yes I think we can use compile_env and this one, bump the requirement to 1.10+, and release 1.7.0?

yordis (approved):

PR #590, opened 2023-05-15
Make EncodeJson/DecodeJson middlewares public

whatyouhide:


lib/tesla/middleware/json.ex
```diff
@@ -61,6 +64,7 @@ defmodule Tesla.Middleware.JSON do
 
   It is used by `Tesla.Middleware.EncodeJson`.
   """
+  @spec encode(Tesla.Env.t(), keyword()) :: Tesla.Env.result()
```
whatyouhide, line 67: Added spec (here and below) since these functions are public.

lib/tesla/middleware/json.ex
```diff
@@ -151,7 +156,15 @@ defmodule Tesla.Middleware.JSON do
...
 
 defmodule Tesla.Middleware.DecodeJson do
-  @moduledoc false
+  @moduledoc """
+  Decodes response body as JSON.
+
+  Only decodes the body if the `Content-Type` header suggests
+  that the body is JSON.
+  """
+  @moduledoc since: "1.8.0"
```
whatyouhide, line 165: Next version.

yordis: The funny part is that I do know people using the module already 😄

yordis (approved):

PR #592, opened 2023-05-25
Explained middleware ordering

LukasKnuth:
I stumbled about this very case previously and had to read into the source code to clear up what was happening.
I understand that I'm somewhat paraphrasing what the actual implementation does (Decoration pattern), but I think for the documentation this is probably okay.

Let me know if the information provided here is sufficient.


yordis (approved):
🚀 💜

PR #596, opened 2023-06-09
Extend a macro-based client with new middleware for runtime usage

alizain:
Usage is pretty simple, helps me re-use a compile-time client with some modifications at runtime.

```elixir
defmodule SomeClient do
    use Tesla

    plug ...
end

# somewhere else
[
    {Tesla.Middleware.BearerAuth, token: 1}
]
|> SomeClient.extend
|> Tesla.get!(...)
```


yordis:
Hey there, do you mind adding unit tests to prove the use case and write documentation about it?

alizain:
@yordis updated, what do you think?

yordis:
See https://github.com/elixir-tesla/tesla/issues/612

PR #603, opened 2023-07-16
Fixes README.md link to Plug Router

lbighetti:
Small fix to Readme link to the Plug Router

It's currently not jumping to the right place as expected as there is a slightly outdated link url. This PR fixes it.


yordis:
💜

PR #606, opened 2023-08-10
feat: improve decompression middleware

### LINKED ISSUE #598: Compression middleware doesn't update `content-*` headers after decompression
When decompressing, `content-encoding` and `content-length` are left unchanged:

```elixir
iex> response = Tesla.client([Tesla.Middleware.Compression]) |> Tesla.get!("http://perdu.com")
iex> response.headers |> List.keyfind("content-encoding", 0)
{"content-encoding", "gzip"}
iex> response.body
"<html><head><title>Vous Etes Perdu ?</title></head><body><h1>Perdu sur l'Internet ?</h1><h2>Pas de panique, on va vous aider</h2><strong><pre>    * <----- vous &ecirc;tes ici</pre></strong></body></html>\n"
iex> response.body |> byte_size()
204
iex> response.headers |> List.keyfind("content-length", 0)  
{"content-length", "163"}
```

Information provided by these headers no longer match the body, and this could users and next middlewares. I believe they should be updated. What do you think?

yordis:
We need to come up with some consensus about the feature!

closes #598

### DIFF
--- a/lib/tesla/middleware/compression.ex
+++ b/lib/tesla/middleware/compression.ex
@@ -25,9 +25,14 @@ defmodule Tesla.Middleware.Compression do
   def call(env, next, opts) do
     env
     |> compress(opts)
-    |> Tesla.put_headers([{"accept-encoding", "gzip, deflate"}])
+    |> add_accept_encoding()
     |> Tesla.run(next)
     |> decompress()
+  end
+
+  @doc false
+  def add_accept_encoding(env) do
+    Tesla.put_headers(env, [{"accept-encoding", "gzip, deflate, identity"}])
   end
 
   defp compressible?(body), do: is_binary(body)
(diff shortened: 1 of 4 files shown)


lib/tesla/middleware/compression.ex
```diff
@@ -61,8 +61,19 @@ defmodule Tesla.Middleware.Compression do
...
+    content_length =
+      body
+      |> byte_size()
+      |> to_string()
+
     env
-    |> Tesla.put_body(decompress_body(env.body, Tesla.get_header(env, "content-encoding")))
+    |> Tesla.put_body(body)
+    # what should happen if the content-encoding is not supported and the body
+    # is not decompressed?
```
yordis, line 74: Need feedback here!

teamon: I think the reasonable thing to do is to keep the headers untouched in case the body was not decompressed. This way one could build a custom middleware and use it alongside Compression to decode various encodings.

Something like

```ex
case decompress_body(env, ...) do
  {:ok, new_body} -> env |> put_body(new_body) |> delete_header(...)
  :unsupported -> env
end
```

yordis:
@teamon give me some review over here 🙏🏻

test/tesla/middleware/compression_test.exs
```diff
@@ -114,7 +116,8 @@ defmodule Tesla.Middleware.CompressionTest do
       {status, headers, body} =
         case env.url do
           "/" ->
-            {200, [{"content-type", "text/plain"}, {"content-encoding", "gzip"}], env.headers}
+            {200, [{"content-type", "text/plain"}, {"content-encoding", "gzip"}],
+             TestSupport.gzip_headers(env)}
```
tanguilp, line 120: Unclear what this function performs

lib/tesla/middleware/compression.ex
```diff
@@ -61,13 +61,30 @@ defmodule Tesla.Middleware.Compression do
...
+
+      body ->
+        put_decompressed_body(env, body)
+    end
+  end
+
+  defp put_decompressed_body(env, body) do
+    content_length =
+      body
+      |> byte_size()
```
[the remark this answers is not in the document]
yordis: Related to https://github.com/wojtekmach/req/issues/239

I am not even sure if we should be adding the length at the end or not, honestly ...

test/tesla/middleware/compression_test.exs
```diff
@@ -69,6 +69,7 @@ defmodule Tesla.Middleware.CompressionTest do
 
   test "decompress response body (gzip)" do
     assert {:ok, env} = CompressionResponseClient.get("/response-gzip")
+    assert env.headers == [{"content-type", "text/plain"}, {"content-length", "17"}]
```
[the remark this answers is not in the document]
yordis: I am OK with it if it doesn't cause indeterministic tests. The whole testing environment needs some clean up, honestly.

test/tesla/middleware/compression_test.exs
yordis: As I said before, the whole testing environment needs some cleanup; it is OK for now.

yordis:
Adding support to a list of decoders ... will take a bit more

lib/tesla/middleware/compression.ex
```diff
@@ -49,6 +51,26 @@ defmodule Tesla.Middleware.Compression do
...
+
+  defmacrop ezstd_loaded? do
+    if Code.ensure_loaded?(:ezstd) do
+      true
+    else
+      quote do
+        Code.ensure_loaded?(:ezstd)
+      end
+    end
+  end
```
teamon, line 72: I'm not sure about this. Wouldn't it be better to provide different middlewares for each compression? 🤔

lib/tesla/middleware/compression.ex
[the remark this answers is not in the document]
wojtekmach: yeah I am almost sure I'll revert the piece that recalculates content-length and drop the header altogether instead.

lib/tesla/middleware/compression.ex
yordis: @teamon Honestly, the whole thing is broken right now 😭 I copied some code from @wojtekmach to make it work 😄 

At some point I am going to ping you to decide on that front, The tricky bit is that if you go for the middleware route, you must configure the middleware in the same order as you need to decode. It is practically impossible to know without the server return from my perspective.

On the flip side, just remove it and only support fewer lists... I honestly copied the code without much thought for now

image

lib/tesla/middleware/compression.ex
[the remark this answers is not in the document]
wojtekmach: Good call, I'll do the same.

lib/tesla/middleware/compression.ex
yordis: I removed the extra encoding for now

yordis:
PR ready

PR #608, opened 2023-08-16
feat: add support for should_retry with arity of 2 to Retry middleware

mdlkxzmcp:
Closes https://github.com/elixir-tesla/tesla/issues/578

I also formatted the middleware.ex as it caused the lining test to fail aaaand made a typo in the branch name 🤦

~I am open to ideas on how to make this work with what is described in that issue... I needed a way to prevent Retry from repeating POST/PUT requests on timeouts, so I wrote it and at first glance thought it was the same issue, but now I see it isn't :)~

Edit: with the addition of the suggested changes by @yordis, this now does in fact close the aforementioned issue 🎊

### DIFF
--- a/lib/tesla/middleware/retry.ex
+++ b/lib/tesla/middleware/retry.ex
@@ -34,6 +34,13 @@ defmodule Tesla.Middleware.Retry do
         {:ok, _} -> false
         {:error, _} -> true
       end
+      # or
+      plug Tesla.Middleware.Retry, should_retry: fn
+        {:ok, %{status: status}}, _env when status in [400, 500] -> true
+        {:ok, _reason}, _env -> false
+        {:error, _reason}, %Tesla.Env{method: :post} -> false
+        {:error, _reason}, _env -> true
+      end
   end
   ```
 
@@ -42,7 +49,8 @@ defmodule Tesla.Middleware.Retry do
   - `:delay` - The base delay in milliseconds (positive integer, defaults to 50)
   - `:max_retries` - maximum number of retries (non-negative integer, defaults to 5)
   - `:max_delay` - maximum delay in milliseconds (positive integer, defaults to 5000)
-  - `:should_retry` - function to determine if request should be retried
+  - `:should_retry` - function with arity of 1 or 2 to determine if request should be retried
+      arity of 1 passes the result, while arity of 2 passes env as second argument (defaults to a match on `{:error, _reason}`)
   - `:jitter_factor` - additive noise proportionality constant
       (float between 0 and 1, defaults to 0.2)
   """
@@ -65,7 +73,7 @@ defmodule Tesla.Middleware.Retry do
       delay: integer_opt!(opts, :delay, 1),
       max_retries: integer_opt!(opts, :max_retries, 0),
       max_delay: integer_opt!(opts, :max_delay, 1),
-      should_retry: Keyword.get(opts, :should_retry, &match?({:error, _}, &1)),
+      should_retry: should_retry_opt!(opts),
       jitter_factor: float_opt!(opts, :jitter_factor, 0, 1)
     }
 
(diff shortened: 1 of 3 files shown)


lib/tesla/middleware/retry.ex
```diff
@@ -84,15 +92,26 @@ defmodule Tesla.Middleware.Retry do
...
-      retry(env, next, context)
-    else
-      res
+    {:arity, should_retry_arity} = :erlang.fun_info(context.should_retry, :arity)
+
+    cond do
+      should_retry_arity == 1 and context.should_retry.(res) ->
+        do_retry(env, next, context)
+
+      should_retry_arity == 2 and context.should_retry.(res, env) ->
```
yordis, line 101:
```suggestion
      should_retry_arity == 2 and context.should_retry.(res, env, context) ->
```

We should also pass the `context,` so the function may be 3 arities now reading the code.

lib/tesla/middleware/retry.ex
```diff
@@ -124,6 +143,19 @@ defmodule Tesla.Middleware.Retry do
     end
   end
 
+  defp should_retry_opt!(opts) do
+    case Keyword.get(opts, :should_retry, &match?({:error, _}, &1)) do
+      should_retry_fun when is_function(should_retry_fun, 1) ->
+        should_retry_fun
```
yordis, line 149: I wonder if we should put a warning log to move away from 1 arity into 3 so there is only one encouraged way, and when breaking changes come, we could remove it.

Thoughts?!

yordis (changes requested):

lib/tesla/middleware/retry.ex
mdlkxzmcp: Pushing the user to transition to the 3-arity version could be quite beneficial since it would force them to think about the cases where this might be desirable. Having the option to keep it simple might be also nice, but imo hinting at customizing the retry logic is a good practice as in many situations you don't want to retry a POST request on a timeout for example.

If you think this is a good direction for the library, then I will add such a warning and modify the code accordingly :)

yordis: Keep them both, we can come back to this later

lib/tesla/middleware/retry.ex
```diff
@@ -42,7 +50,9 @@ defmodule Tesla.Middleware.Retry do
   - `:delay` - The base delay in milliseconds (positive integer, defaults to 50)
   - `:max_retries` - maximum number of retries (non-negative integer, defaults to 5)
   - `:max_delay` - maximum delay in milliseconds (positive integer, defaults to 5000)
-  - `:should_retry` - function to determine if request should be retried
+  - `:should_retry` - function with an arity of 1 or 3 used to determine if the request should
+      be retried the first argument is the result, the second is the env and the third is
+      the context: options + `:retries` (defaults to a match on `{:error, _reason}`)
```
yordis, line 55: the best way to document such complexity is by having a `@type` maybe called `should_retry_func` and document it there.

You do not have to do it, just commeting so that in the future others could find this comment.

yordis (approved):

yordis:
@teamon going to wait for your feedback just in case, but thus far, I am OK with the addition.

lib/tesla/middleware/retry.ex
mdlkxzmcp: I wanted to do something like that, but I haven't seen anything of the sort in any of the other middlewares... If you want, I can make a follow-up PR that documents all the options. The question is should I leverage typedocs and remove the `## Options` section in favor of using something like:

```elixir
@typedoc "Supported options."
@type options :: %{optional(:delay) => delay(), ...}

@typedoc "The base delay in milliseconds (positive integer, defaults to 50)"
@type delay :: pos_integer()
```

And then just reference it with `t:options/0` in the moduledoc?

yordis: I tend to say yes, lets give it a try

PR #613, opened 2023-09-10
fix: merging adapter options using Opts middleware

### LINKED ISSUE #367: Adapter options aren't merged correctly

yordis:
fixes #367

### DIFF
--- a/lib/tesla/middleware/opts.ex
+++ b/lib/tesla/middleware/opts.ex
@@ -17,6 +17,16 @@ defmodule Tesla.Middleware.Opts do
 
   @impl Tesla.Middleware
   def call(env, next, opts) do
-    Tesla.run(%{env | opts: env.opts ++ opts}, next)
+    adapter =
+      env.opts
+      |> Keyword.get(:adapter, [])
+      |> Keyword.merge(opts[:adapter] || [])
+
+    opts =
+      env.opts
+      |> Keyword.merge(opts)
+      |> Keyword.put(:adapter, adapter)
+
+    Tesla.run(%{env | opts: opts}, next)
   end
 end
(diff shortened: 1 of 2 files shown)


yordis:
@teamon share some love over here 🙏🏻

yordis:
related to https://github.com/elixir-tesla/tesla/pull/500

Issue #612, opened 2023-09-10
[Proposal] Lower the usage of macros


yordis:
> https://github.com/elixir-tesla/tesla/issues/367#issuecomment-610342739
> I would discourage usage of macros for DRY. It's better to use the [Runtime middleware](https://github.com/teamon/tesla#runtime-middleware).

After many years of using Tesla, I came across too many situations that are much simpler to deal with using the Runtime middleware, slowing questioning the value of having macros.

@teamon I am not sure if you still feel strongly about having such a feature or if it would be something that maybe is worth deprecating/discouraging/removing in the future.

yordis:
https://github.com/elixir-tesla/tesla/pull/575/files is an example where the macro style ends up wanting a Runtime Middleware style of execution (not quite, but similar)

So eventually, every middleware will end up having to support `MFA,` static values, and function pointers ... it is just a matter of somebody wanting to do it their style.

Versus Runtime Middleware removing all that complexity.

teamon:
[🎶 Hello macros my old friend 🎶 ](https://github.com/elixir-tesla/tesla/issues/377)

The first step would be to update the README and make the runtime middleware the default way, while keeping macros for backwards compatibility.

yordis:
Docs updated for now, v2 we **may** remove the macro all together

PR #621, opened 2023-10-01
chore: add warning message to Tesla.Builder

yordis:
related to #612

Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

### DIFF
--- a/lib/tesla/builder.ex
+++ b/lib/tesla/builder.ex
@@ -5,6 +5,19 @@ defmodule Tesla.Builder do
   @body ~w(post put patch)a
 
   defmacro __using__(opts \\ []) do
+    if not Application.get_env(:tesla, :disable_deprecated_builder_warning, false) do
+      IO.warn("""
+      `use Tesla.Builder` and `use Tesla` are soft-deprecated. It will be removed in future major version in favor of
+      Runtime Configuration instead. Please share your feedback at https://github.com/elixir-tesla/tesla/discussions/732
+
+      If you can turn off this warning, add the following to your config.exs:
+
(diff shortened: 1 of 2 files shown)


yordis:
@teamon, I need your approval over here. I intend to follow up by writing a proper documentation guide regarding migrating, but I would like us to send an awareness message sooner rather than later.

This is an annoying message but simple to turn off.

yordis:
@teamon do you have any strong reason against me adding the log? I don't know what else to do to make thing safer. Adding it for at least 1 year and then remove it is what I intent.

I also need to point people to https://hexdocs.pm/tesla/v1-macro-migration.html

lib/tesla/builder.ex
```diff
@@ -2,7 +2,26 @@ defmodule Tesla.Builder do
   @http_verbs ~w(head get delete trace options post put patch)a
   @body ~w(post put patch)a
 
+  require Logger
+
   defmacro __using__(opts \\ []) do
+    unless Application.get_env(:tesla, :warning_acknowledged, false) do
+      Logger.warn("""
+      `use Tesla.Builder` and `use Tesla` are soft-deprecated. It will be removed in future major version in favor of
```
adamu, line 10: How about [`IO.warn`](https://hexdocs.pm/elixir/IO.html#warn/1) to avoid the dependency? Also, I think warning + removal is a regular deprecation, not a soft one.

```suggestion
      IO.warn("""
      `use Tesla.Builder` and `use Tesla` are deprecated. It will be removed in future major version in favor of
```

PR #622, opened 2023-10-08
feat: allow to configure the prefix for the telemtry events

yordis:
Recently, @seancribbs asked to be able to control the prefix.

What is the proper solution here?

1. Allow people to customize the prefix
2. Hook into the events and, based on the data, do something with it, such as using the HOST information of the request or User-Agent
3. ?????

### DIFF
--- a/lib/tesla/middleware/telemetry.ex
+++ b/lib/tesla/middleware/telemetry.ex
@@ -25,6 +25,7 @@ if Code.ensure_loaded?(:telemetry) do
     ## Options
 
     - `:metadata` - additional metadata passed to telemetry events
+    - `:prefix` - prefix for telemetry events. Defaults to `:tesla`
 
     ## Telemetry Events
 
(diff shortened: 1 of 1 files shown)


bryannaegele:
Other libraries allow this (plug.telemetry, ecto, etc). 👍🏻

teamon:
@yordis Please take a look https://github.com/elixir-tesla/tesla/pull/390#issuecomment-644243214

bryannaegele:
Yeah, I misspoke here in this PR. Sorry. They can key off something in the metadata. We do need the event name to be consistent.

If it isn't already possible, if the user can apply some metadata to the event then it could be trivial to figure out which client it's coming from.

teamon:
That's why I'm referencing this comment, I don't see the need to change the prefix (and IMO it will cause more harm than good)

bryannaegele:
@yordis can you please revert?

yordis:
oooopppppsss sorry! 😭 😭 😭 😭 😭 😭

PR #625, opened 2023-10-10
Add support for Gun 2.0 - Feat: update hackney

carrascoacd:
Solves the compatibility problems detected on Gun 2.0 https://github.com/elixir-tesla/tesla/pull/587 cc/ @yordis


test/tesla/adapter/gun_test.exs
```diff
@@ -186,14 +186,14 @@ defmodule Tesla.Adapter.GunTest do
     assert response.status == 500
   end
 
-  test "error on socks proxy" do
+  test "error waiting for response on socks proxy" do
```
carrascoacd, line 189: @yordis maybe we just can get rid of this test since socks proxy is already supported.

yordis: I do not see why not, but what about `v1` people? I am just wondering if this will impose a major Tesla version bump.

yordis: https://github.com/elixir-tesla/tesla/pull/587#issuecomment-1702687962

nevermind

yordis:
What is the minimal required version of `:gun` dep?

test/tesla/adapter/gun_test.exs
carrascoacd: Oh you are right about the v1. In my opinion, I'd go for enforcing >= 2.0 and release a major version of tesla.

yordis: 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭 😭

teamon: If we only change the error returned I don’t think we need a major version bump, let’s just make sure all tests pass for both versions of gun (we can match on {:error, _})

carrascoacd: But @teamon this is not the only change, the [`headers`](https://ninenines.eu/docs/en/gun/2.0/manual/gun.headers/) fn is not available in 1.0 so we will need to add an `if` use one function or other depending on the version. What about the tests here, how could you ensure that it runs against two versions?

teamon: That's not a big problem:

```ex

if version_check do
  def dosmth(...), :gun.headers(...)
else
  def dosmth(...), :gun.other_fun(...)
end
```

yordis: Welcome to the most beautiful feature of Elixir 😄 😄 😄 😄

yordis: ..... right 😭

carrascoacd: Oh! I see si you rely on a test mix.lock file, dark magic 🧙 . Cool, I'll do it in that way 😀

carrascoacd: Ok guys. Please take a look again. I thought about defining two modules based on the version and duplicating the whole files but, from now, I think that we can keep it like that as the differences are very small.

Notice that I'm doing the check more explicitly, using `Version` instead of `function_exported?`.

carrascoacd: PS: please, could you re-run the failed tests? I think they are flaky ones [test/tesla/middleware/retry_test.exs].

mix.exs
```diff
@@ -85,7 +85,7 @@ defmodule Tesla.Mixfile do
 
       # httparrot dependencies
       {:httparrot, "~> 1.3", only: :test},
-      {:cowlib, "~> 2.9", only: :test, override: true},
+      {:cowlib, "~> 2.9", only: [:dev, :test], override: true},
```
yordis, line 88: this used to drive me crazy!

.github/workflows/test.yml
```diff
@@ -38,6 +38,44 @@ jobs:
...
+      uses: erlef/setup-beam@v1
+      with:
+        elixir-version: ${{ matrix.elixir }}
+        otp-version: ${{ matrix.otp }}
+        version-type: strict
+    - name: Restore dependencies cache
+      uses: actions/cache@v3
+      with:
+        path: deps
+        key: ${{ runner.os }}-mix-${{ hashFiles('**/mix.lock') }}
```
yordis, line 64:
```suggestion
        key: ${{ runner.os }}-mix-${{ hashFiles('test/lockfiles/gun1.lock') }}
```

Wouldn't be that lock file?

carrascoacd: Retry test failing again. I'll take a look into it, tests are not isolated.

yordis (approved):

yordis:
@carrascoacd I just made a mistake ... do you mind opening the PR against `master` instead of the dependabot branch instead?

We can close the `dependabot/hex/gun-2.0.1` and use yours! I am sorry!

yordis:
https://github.com/elixir-tesla/tesla/pull/630 I fixed at https://github.com/elixir-tesla/tesla/pull/630

Your contributions are there!

My apologies for the mess, I realized the target branch wasn't `master` after I sent the wrong message to dependabot ... sorry!

carrascoacd:
Thanks! Nothing to sorry about.

PR #626, opened 2023-10-10
feat: add ssl opt to httpc by default

### LINKED ISSUE #293: Secure by default
# The Motivation

Since my earlier days as developer I always have been astonished with the mindset of our industry of being insecure by default, where normally security is opt-in, instead of opt-out, and this mindset is one of the main culprits for so many data-breaches that occur every week/month/year. 

So the current mindset is to release software with all the doors and windows of your home open, while your are away, and then hope that the developers using it will learn how to properly  close all that doors and windows, and then also hope that when they learn how to properly do it, that they don't forget one open.

Just to put things in perspective, try to use [shodan.io](https://www.shodan.io/search?query=mongodb) to see how software ends up in production without being properly secured, due to the current mindset of our industry of preferring convenience over security.

This all started in [this tweet](https://twitter.com/Exadra37/status/1118874708064256001) and I was asked in [this other tweet](https://twitter.com/iteamon/status/1119254174326300673) to open this issue.   

# The Context

So this have been brought up before by @lau in the issue #235 and by @rawkode in issue #138.

By @lau

> The default adapter is httpc, which does not check certificates when using HTTPS. This seems like an unsafe default. Many people will probably use the default and use it to communicate with APIs. Perhaps with sensitive data.

I cannot agree more with this, but the issue on the time was solved with only a note in the README.

Alerts, notes in the README or any other docs will be missed, ignored or forgotten, by beginners, juniors and senior developers, because human beings tend to prefer convenience above anything else., but convenience should not be put in front of Security.

# The Proposal

@teamon says on [this comment](https://github.com/teamon/tesla/issues/138#issuecomment-350702593):

> The point of Tesla is not to get rid of hackney, nor any other http adapter. The point is to let the end use choose the adapter that suits best or write a custom one if needed.

So in the README we have:

> Configure default adapter in config/config.exs (optional).

 But instead we should have:

> Configure default adapter in config/config.exs (**REQUIRED**).

And we could then have a list of possible secure by default adapters.

This change would leave Tesla agnostic of an adapter as per the goal of @teamon, but would not leave Tesla insecure by default... The developer would be required to explicitly opt-in for the adapter to use.

So this means that if the Tesla was to be used without configuring the adapter, an exception should be raised with a very clear message.

Once this is a breaking change a major version would be needed.

> **PS:** I appeal to all developers to always embrace **Secure By Default** aproach when building software.

yordis:
fixes #293

Signed-off-by: Yordis Prieto <yordis.prieto@gmail.com>

### DIFF
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -18,7 +18,7 @@ defmodule Tesla.Adapter.Httpc do
   @impl Tesla.Adapter
   def call(env, opts) do
     opts = Tesla.Adapter.opts(@override_defaults, env, opts)
-    opts = Tesla.Adapter.opts(default_ssl_opt(), env, opts)
+#    opts = Tesla.Adapter.opts(default_ssl_opt(), env, opts)
 
     with {:ok, {status, headers, body}} <- request(env, opts) do
       {:ok, format_response(env, status, headers, body)}
@@ -32,6 +32,7 @@ defmodule Tesla.Adapter.Httpc do
   defp default_ssl_opt do
     # TODO: verify that requires OTP 25+
     # TODO: verify that does not require any Elixir version
+    # TODO: maybe use Castore for now? cacertfile: CAStore.file_path(),
     [
       verify: :verify_peer,
       cacerts: :public_key.cacerts_get(),
--- a/lib/tesla/adapter/httpc.ex
+++ b/lib/tesla/adapter/httpc.ex
@@ -18,6 +18,7 @@ defmodule Tesla.Adapter.Httpc do
   @impl Tesla.Adapter
   def call(env, opts) do
     opts = Tesla.Adapter.opts(@override_defaults, env, opts)
+    opts = Tesla.Adapter.opts(default_ssl_opt(), env, opts)
 
     with {:ok, {status, headers, body}} <- request(env, opts) do
       {:ok, format_response(env, status, headers, body)}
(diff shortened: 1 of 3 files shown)


lib/tesla/adapter/httpc.ex
```diff
@@ -18,12 +20,38 @@ defmodule Tesla.Adapter.Httpc do
...
+          ]
+        ]
+      ]
+      Tesla.Adapter.opts(default_ssl_opt, env, opts)
+    end
+  else
+    def add_default_ssl_opt(env, _opts) do
+      env
+    end
+  end
```
maennchen, line 53: As written in Slack

```suggestion
  cond do
    # TODO: remove this once OTP 25+ is required
    current_otp_version <= 25 ->
      def add_default_ssl_opt(env, _opts) do
        # TODO: maybe use Castore for now? cacertfile: CAStore.file_path(),
      
        # https://github.com/erlang/otp/blob/33d99c5f50988c4bf124974ce3340bf18dbf0d08/lib/inets/src/http_client/httpc.erl#L461-L476
        default_ssl_opt = [
          ssl: [
            verify: :verify_peer,
            cacerts: :public_key.cacerts_get(),
            customize_hostname_check: [
              match_fun: :public_key.pkix_verify_hostname_match_fun(:https)
            ]
          ]
        ]
        Tesla.Adapter.opts(default_ssl_opt, env, opts)
      end
      
    # TODO: remove this once OTP 25+ is required
    current_otp_version == 25 ->
      def add_default_ssl_opt(env, _opts) do
        default_ssl_opts = :httpc.ssl_verify_host_options(true)
        Tesla.Adapter.opts(default_ssl_opt, env, opts)
      end
      
    current_otp_version >= 26 ->
      def add_default_ssl_opt(env, _opts), do: env
    end
```

yordis:
@maennchen do you know what am I am doing wrong that the SSL is not returning an error, but logging it

> Notice:  badssl expired.badssl.com [L#52]2024-07-01 06:12:48.276 [notice] TLS :client: In state :certify at ssl_handshake.erl:2109 generated CLIENT ALERT: Fatal - Certificate Expired

yordis:
@teamon @maennchen ready to code review 🙏🏻

teamon (approved):
Looks reasonable to me 🙇‍♂️

PR #627, opened 2023-10-18
Update Tesla.Middleware.Logger for Elixir 1.11+

halostatue:
Elixir 1.15 is now warning whenever the `:warn` level is used with the
recommendation of using `:warning` instead. The `Logger` middleware uses
`Logger.log/3`, so a compile-time warning would not be shown, only
a runtime warning.

This change adds `:warning` as a valid configuration option for Tesla
log level and automatically translates `:warn` or `:warning` to the
appropriate level (`:warn` for Elixir 1.10.x, `:warning` for Elixir
1.11.x or higher, as `Logger.warning/2` was introduced in 1.11.0).


yordis:
@teamon how do you feel about requiring `elixir 1.11`?

Issue #566, opened 2023-03-31
Middleware.PathParams  :erlang.binary_to_existing_atom error for google APIs


hiagomeels:
Middleware.PathParams  :erlang.binary_to_existing_atom error for google APIs

I'm migrating from another lib to Tesla and I found an issue while trying to use got integrated [this API](https://developers.google.com/business-communications/business-messages/guides/how-to/launch/agents?method=api#launch-agent).

This is the error that I got:

```elixir
** (ArgumentError) argument error
     code: @middleware.call(%Env{url: "/v1/brands/:brand_id/agents/:agent_id:requestLaunch", opts: opts}, [], nil)
     stacktrace:
       :erlang.binary_to_existing_atom("requestLaunch", :utf8)
       (tesla) lib/tesla/middleware/path_params.ex:39: anonymous fn/3 in Tesla.Middleware.PathParams.build_url/2
       (elixir) lib/regex.ex:731: Regex.apply_list/5
       (elixir) lib/regex.ex:732: Regex.apply_list/5
       (elixir) lib/regex.ex:726: Regex.apply_list/5
       (elixir) lib/regex.ex:668: Regex.do_replace/4
       (tesla) lib/tesla/middleware/path_params.ex:31: Tesla.Middleware.PathParams.call/3
       test/tesla/middleware/path_params_test.exs:91: (test)
```

The problem is that the path for this API request is using a colon to indicate one action in the path `/v1/brands/__BRAND_ID__/agents/__AGENT_ID__:requestLaunch` (see doc).

I know that I could just pass the ":requestLaunch" as a path param as well:

```elixir
path = "/v1/brands/:brand_id/agents/:agent_id:requestLaunch"
path_params = [
  agent_id: ...,
  brand_id: ...,
  requestLaunch:  ":requestLaunch" 
]
...
```

But it is very odd/ugly, and I think we should fix it on the middleware.

I have the following suggestion that we could:

## 1 - Add a rescue clause
Since we are using the `String.to_existing_atom/1` function on `lib/tesla/middleware/path_params.ex:39` we could rescue it if it fails.

```elixir
defmodule Tesla.Middleware.PathParams do
  ...
  defp build_url(url, params) do
    Regex.replace(@rx, url, fn match, key ->
      try do
        to_string(params[String.to_existing_atom(key)] || match)
      rescue
        ArgumentError -> match
      end
    end)
  end
end
```

_I'm unsure if we will want it not to throw an error if the param does not exist._

[...]

PR #628, opened 2023-10-18
feat: Middleware.PathParams: add OpenAPI params

halostatue:
Tesla.Middleware.PathParams has been extended to support OpenAPI style parameters in addition to the existing Phoenix style parameters.

- Phoenix parameters begin with `:` and a letter (`a-zA-Z`) and continue to the next non-word character (not `a-zA-Z0-9_`). Examples include `:id`, `:post_id`, `:idPost`.

- OpenAPI parameters are contained in braces (`{}`), must begin with a letter (`a-zA-Z`) and may contain word characters and hyphens (`-a-zA-Z0-9_`). Examples inlucde `{id}`, `{post_id}`, `{IdPost}`.

Parameter value objects may be provided as maps with atom or string keys, keyword lists, or structs. During path resolution, to avoid potential atom exhaustion, the parameter value object is transformed to a string-keyed map.

Parameter values that are `nil` or that are not present in the value object are not replaced; all other values are converted to string (with `to_string`) and encoded with `application/x-www-form-urlencoded` formatting. If the value does not implement the `String.Chars` protocol, it is the caller's responsibility to convert it to a string prior to passing it to the Tesla client using this middleware.

Execution time is determined by the number of parameter patterns in the path.

Closes: #566

### DIFF
--- a/lib/tesla/middleware/path_params.ex
+++ b/lib/tesla/middleware/path_params.ex
@@ -1,12 +1,33 @@
 defmodule Tesla.Middleware.PathParams do
   @moduledoc """
-  Use templated URLs with separate params.
+  Use templated URLs with provided parameters in either Phoenix style (`:id`)
+  or OpenAPI style (`{id}`).
 
-  Useful when logging or reporting metric per URL.
+  Useful when logging or reporting metrics per URL.
+
+  ## Parameter Values
+
+  Parameter values may be `t:struct/0` or must implement the `Enumerable`
+  protocol and produce `{key, value}` tuples when enumerated.
+
+  ## Parameter Name Restrictions
+
+  Phoenix style parameters may contain letters, numbers, or underscores,
+  matching this regular expression:
+
+    :[a-zA-Z][_a-zA-Z0-9]*\b
+
+  OpenAPI style parameters may contain letters, numbers, underscores, or
+  hyphens (`-`), matching this regular expression:
+
+    \{[a-zA-Z][-_a-zA-Z0-9]*\}
+
+  In either case, parameters that begin with underscores (`_`), hyphens (`-`),
+  or numbers (`0-9`) are ignored and left as-is.
 
   ## Examples
 
-  ```
+  ```elixir
   defmodule MyClient do
     use Tesla
 
@@ -16,7 +37,12 @@ defmodule Tesla.Middleware.PathParams do
 
     def user(id) do
       params = [id: id]
-      get("/users/:id", opts: [path_params: params])
+      get("/users/{id}", opts: [path_params: params])
+    end
+
+    def posts(id, post_id) do
+      params = [id: id, post_id: post_id]
+      get("/users/:id/posts/:post_id", opts: [path_params: params])
     end
   end
   ```
@@ -24,19 +50,36 @@ defmodule Tesla.Middleware.PathParams do
 
   @behaviour Tesla.Middleware
 
-  @rx ~r/:([a-zA-Z]{1}[\w_]*)/
-
   @impl Tesla.Middleware
   def call(env, next, _) do
     url = build_url(env.url, env.opts[:path_params])
     Tesla.run(%{env | url: url}, next)
   end
 
+  @rx ~r/:([a-zA-Z][a-zA-Z0-9_]*)|[{]([a-zA-Z][-a-zA-Z0-9_]*)[}]/
+
   defp build_url(url, nil), do: url
 
-  defp build_url(url, params) do
-    Regex.replace(@rx, url, fn match, key ->
-      to_string(params[String.to_existing_atom(key)] || match)
+  defp build_url(url, params) when is_struct(params), do: build_url(url, Map.from_struct(params))
+
+  defp build_url(url, params) when is_map(params) or is_list(params) do
+    safe_params = Map.new(params, fn {name, value} -> {to_string(name), value} end)
+
+    Regex.replace(@rx, url, fn
+      # OpenAPI matches
+      match, "", name -> replace_param(safe_params, name, match)
+      # Phoenix matches
+      match, name, _ -> replace_param(safe_params, name, match)
     end)
+  end
+
+  defp build_url(url, _params), do: url
+
+  defp replace_param(params, name, match) do
+    case Map.fetch(params, name) do
+      {:ok, nil} -> match
+      :error -> match
+      {:ok, value} -> URI.encode_www_form(to_string(value))
+    end
   end
 end
--- a/test/tesla/middleware/path_params_test.exs
+++ b/test/tesla/middleware/path_params_test.exs
@@ -1,86 +1,303 @@
 defmodule Tesla.Middleware.PathParamsTest do
   use ExUnit.Case, async: true
+
   alias Tesla.Env
 
   @middleware Tesla.Middleware.PathParams
 
-  test "no params" do
-    assert {:ok, env} = @middleware.call(%Env{url: "/users/:id"}, [], nil)
-    assert env.url == "/users/:id"
+  defmodule TestUser do
+    defstruct [:id]
   end
 
-  test "passed params" do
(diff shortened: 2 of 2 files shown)


halostatue:
> Related to #566 as well
> 
> > This does not handle "nested" OpenAPI-style parameters well (that is, {post{id}} or {id{post}}), and I'm not sure what the behaviour here should be.
> 
> I couldn't find docs about that in the Open API spec. Do you mind linking where is this possible? I didn't know that. That situation seems weird anyway, most people would flat that to `post_id`

I don't believe it is legal, but the implementation in this PR is naïve and will handle it oddly. `"{post{id}}"` with `[post: 3, id: 2]` will result in `"{post2}"` whereas `"{id{post}}"` will result in `"{id3}"`. This gets particularly ugly if you do the latter with `[post: 3, id: 2, id3: 5]`, which will result in `"5"`. It's been a little while since I wrote the code and the PR message, but I believe I was trying to point out limitations.

> > This mostly follows the Phoenix implementation for path parameters, excepting that capital letters are permitted. This matches the previous version's implementation.
> > Try to avoid breaking changes, we can always document the ideal implementation and wait until we have the opportunity to do so.
> 
> > I would personally want to have different parameter identifier matching rules for :id vs {id} formats, but that gets complicated.
> 
> I would like `{name}` to stick to whatever Open API spec says.

That is [ugly](https://swagger.io/specification/#path-templating). 'The value for these path parameters MUST NOT contain any unescaped "generic syntax" characters described by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3): forward slashes (/), question marks (?), or hashes (#).'

In other words, I think that `{0}` would be permitted.

> > This does not handle "nested" OpenAPI-style parameters well (that is, {post{id}} or {id{post}}), and I'm not sure what the behaviour here should be. Strictly speaking, we should probably be doing this somewhat differently than the logic currently implemented or the previous regex-based logic:
> 
> It is OK we have a working version without "nesting", and yes, we should move away from regex if possible, I like your proposal for the macro 🤯

The macro approach has definite downsides, *especially* as the macro-style of Tesla client declaration is disfavoured (I am slowly trying to convert everything to runtime configured) and it would require Tesla's URL handling to be aware of such an iodata-ish approach.

yordis:
Keep going, get to a place where you are happy with things, open for code review, and we open up the discussion to figure out what should be the end-state.

Thank you so much for the contribution thus far!

halostatue:
I’ll do a quick once over this weekend; we are currently using a *variant* of this in production and it works very nicely, as long as you don't do silly things.

halostatue:
Sorry for the delays, but this is now ready. There were a couple of issues we found in practice that I put back in place.

lib/tesla/middleware/path_params.ex
```diff
@@ -34,9 +54,50 @@ defmodule Tesla.Middleware.PathParams do
...
 
+  defp build_url(url, %_{} = params), do: build_url(url, Map.from_struct(params))
+
   defp build_url(url, params) do
-    Regex.replace(@rx, url, fn match, key ->
-      to_string(params[String.to_existing_atom(key)] || match)
-    end)
+    params
+    |> Enum.map(fn {name, value} -> {to_string(name), value} end)
+    |> Enum.sort_by(fn {name, _} -> String.length(name) end, :desc)
```
yordis, line 62: I am a bit confused by the need of sorting 🤔

lib/tesla/middleware/path_params.ex
```diff
@@ -34,9 +54,50 @@ defmodule Tesla.Middleware.PathParams do
...
+    |> Enum.reduce(url, &replace_parameter/2)
   end
+
+  # Do not replace parameters with nil values.
+  defp replace_parameter({_name, nil}, url), do: url
+  defp replace_parameter({name, value}, url), do: replace_parameter(to_string(name), value, url)
+
+  defp replace_parameter(<<h, t::binary>>, value, url)
+       when h in ?a..?z or h in ?A..?Z do
```
yordis, line 71: I wonder if we should care at all in the end 🤔

lib/tesla/middleware/path_params.ex
```diff
@@ -34,9 +54,50 @@ defmodule Tesla.Middleware.PathParams do
...
+
+      name ->
+        encoded_value =
+          value
+          |> to_string()
+          |> URI.encode_www_form()
+
+        url
+        |> String.replace(":#{name}", encoded_value)
+        |> String.replace("{#{name}}", encoded_value)
```
yordis, line 84: I wonder if we should pick one or another based on a configuration. My stand is to wait until somebody complain being "slow", but I would like to hear your thoughts about it

lib/tesla/middleware/path_params.ex
halostatue: Because this maintains Phoenix-style parameters, and we need to support `/user/:id/post/:id_post` and replacing in reverse order ensures that `[id_post: 3, id: 1]` works. There *may* be an issue when you just put `[id: 1]` (I think that might do `/user/1/post/1_post`; I will need to add a test and figure out how to fix it.)

yordis (commented):
It looks great so far, I left couple questions about it.

lib/tesla/middleware/path_params.ex
halostatue: Yeah. That's going to be a problem. I need to figure out how to address this particular issue. I think that I’m going to need to have separate steps for handling Phoenix-style parameters vs OpenAPI style parameters.

yordis: I would branch between OpenAPI and Phoenix-style as soon as we can, and figure out the complexity independently.

I am OK copying Phoenix source code, but as long as OpenAPI is outside of the complexity.

halostatue:
I have a better fix in a second commit and will squash before merging.

lib/tesla/middleware/path_params.ex
halostatue: This was put in part because of the previous restrictions. For phoenix-style parameters, we absolutely care, IMO. For OpenAPI parameters…less so, but I still think that the behaviours documented are good.

lib/tesla/middleware/path_params.ex
halostatue: If we wanted to separate it — and I’m not sure with the most recent commit that it is necessary — I would have something like `OpenApiPathParams` and leave phoenix style parameters to `PathParam`. I would probably keep a mostly-common implementation because this *also* fixes #566.

yordis: I would lean towards adding `OpenApiPathParams` module honestly! I am curious to see how simple could be based on its constraints

halostatue: Hm. The main difference would be instead of using one regex (`~r/PHOENIX_STYLE|OPENAPI_STYLE/`) there would be separate regexes. The `Regex.replace/4` call would use the appropriate regex for its type and instead of having to know `match, "", openapi` and `match, phoenix, _openapi`, we would just be doing `match,  name`.

So not *that* much simpler than what we have right now.

yordis:
Anything last words? :shipit: ?

halostatue:
We shouldn't see anything much worse in performance because the current solution uses a similar `Regex.replace`. I’m going to squash it down and include a reference that this closes #566 (because it does!). I will reword the unified commit message for the next release notes.

halostatue:
Squashed and main PR message ready to go.

I think that having a mechanism to tokenize URLs / URL paths and pass that to Tesla is a slightly larger discussion because it is more invasive and may have negative impacts on *other* plug-ins if not supported. Probably better for a 2.0 discussion, but it would make specific path matching easier and more reliable, IMO.

Although the code here is *different* than what we're using in production, it’s somewhat closer to the original implementation and has nothing but upsides, IMO.

yordis (approved):

yordis:
🚀 💜

PR #639, opened 2023-12-12
feat: support retry-after header in retry middleware

matt-brale-xyz:
Adds :use_retry_after_header option to allow the client to align its retry strategy with server-side rate limiting


matt-brale-xyz:
Let me know if there's any feedback you have on the proposed change - thanks!

mix.exs
```diff
@@ -50,6 +50,7 @@ defmodule Tesla.Mixfile do
   defp deps do
     [
       {:mime, "~> 1.0 or ~> 2.0"},
+      {:date_time_parser, "~> 1.2.0"},
```
yordis, line 53: Could this be done without this package? Do you mind sharing the needs for it?

yordis: I created an issue in Elixir Lang Core Google Group: https://groups.google.com/g/elixir-lang-core/c/FYVQP44UPoI

Honestly, this sounds such important parsing to have, but I rather keep it simple for now

yordis: https://github.com/elixir-plug/plug/blob/cc535b13c5569f199dbde5ca5c030eb8a6d61e8b/lib/plug/conn/cookies.ex#L99:L139

matt-brale-xyz: Sounds good - I'll add some parsing code and take out the dependency

matt-brale-xyz: Adapted the implementation from there and added some extra tests to make sure I didn't break it when I changed the error handling.

yordis:
🚀 💜

PR #640, opened 2023-12-19
feat: override adapter at OTP app level

c4710n:
Thanks for all your hard works, Tesla already [has great support on choosing adapter at different levels](https://github.com/elixir-tesla/tesla/blob/189ffa5b0c26fcce1a6c88e4fe2fb17de772e578/lib/tesla.ex#L123).

But, it still have one case that tesla doesn't support - override adapter at OTP app level.

This PR adds that feature, which is useful when we want to use different adapter for different OTP apps.

---
An example - I installed two packages from Hex:
+ `foo`, whose OTP app name is `:foo`, the underlying HTTP client is **Tesla**
+ `bar`, whose OTP app name is `:bar`, the underlying HTTP client is **Tesla**, too.

I want to use `:finch` as the adapter for them all, but due to some reason, I have to set HTTP proxy for `bar`. But, `:finch` doesn't support proxy, for now. So I have to use `:hackney`, which supports proxy.

With this PR, I can do something like:

```
config :telsa, :adapter, {Tesla.Adapter.Finch, name: MyFinch}
config :tesla, :bar, adapter: Tesla.Adapter.Hackney # override adapter for :bar in an easy way.
```

---

Feel free to comment. I can make any modifications necessary to ensure this PR is mergeable. ;)

### DIFF
--- a/lib/tesla.ex
+++ b/lib/tesla.ex
@@ -123,6 +123,7 @@ defmodule Tesla do
   def effective_adapter(module, client \\ %Tesla.Client{}) do
     with nil <- client.adapter,
          nil <- adapter_per_module_from_config(module),
+         nil <- adapter_per_otp_app_from_config(module),
          nil <- adapter_per_module(module),
          nil <- adapter_from_config() do
       adapter_default()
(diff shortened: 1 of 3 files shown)


yordis:
Hey there, sorry it took me this long to get to you; in the latest Tesla version, I removed all documents related to macros to discourage the use of macros without removal.
That means that you can pass any adapter using the strategy you wish to `Tesla.client/2`

I will close the PR; if some folks out there have some serious issue in production that such a feature is granted, please let me know; otherwise, I rather avoid adding even more features that may cause even more work for a v2 breaking change moment.

Thank you so much for your work regardless!

PR #649, opened 2024-01-09
Update tesla.ex

bradhanks:
warning: using map.field notation (without parentheses) to invoke function Tesla.__middleware__() is deprecated, you must add parentheses instead: remote.function()


yordis (approved):

yordis:
🚀 💜

PR #670, opened 2024-04-15
Update README to use newest versions of dependencies

arcanemachine:
- Also prefer the `~>` operator when installing packages (e.g. Jason)
  - This is the format used in [the Jason documentation](https://github.com/michalmuskala/jason#installation)

- Pin to major versions instead of minor


yordis (approved):

yordis:
🚀 💜

PR #675, opened 2024-05-09
remove seemingly unnecessary step

byronanderson:
This will bring a stream into memory, which can take a lot of memory unnecessarily

Admittedly there is a lot of intermediate code that I can't attach here, but replacing the finch adapter with the httpc adapter looks like this with a 10G multipart request body:

Finch adapter:

finch

httpc adapter:

httpc

(Notice that the units are different between the two screenshots, one is in MB and the other in GB)

I can try to characterize the memory issue I am seeing with a test if you would like to see, but uploading a big file to see a memory spike in a test seems like a test you wouldn't want in the long term.


yordis:
🚀 💜

PR #682, opened 2024-06-18
Update CI with current Elixir and Erlang/OTP versions

nathany-copia:
> I am trying to maintain (in CI) the latest 2 major versions, so Elixir 15+ and OTP 25+; wait 2-3 months and it will be Elixir 16+ and OTP 26+

https://github.com/elixir-tesla/tesla/pull/623#issuecomment-2176886751

This approach runs fewer permutations, so we're not attempting an old version of Elixir on a new version of Erlang. We also can flag the latest version or release candidate as experimental, so that it doesn't block merging, but still reports the failures.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,11 +13,11 @@ jobs:
           - otp: '25.3.2.12'
             elixir: '1.15.8'
             experimental: false
-            lint: true
+            lint: false
           - otp: '26.2.5'
             elixir: '1.16.3'
             experimental: true
-            lint: false
+            lint: true
           - otp: '27.0'
             elixir: '1.17.1'
             experimental: true
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,14 +6,22 @@ jobs:
   Test:
     runs-on: ubuntu-latest
     name: OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}}
+    continue-on-error: ${{ matrix.experimental }}
     strategy:
       matrix:
-        elixir:
-          - 1.14
-          - 1.13
-        otp:
-          - 25.3
-          - 24.3
+        include:
+          - otp: '25.3'
+            elixir: '1.15'
+            experimental: false
+          - otp: '26.2'
+            elixir: '1.16'
+            experimental: false
+          - otp: '26.2'
+            elixir: '1.17'
+            experimental: true
+          - otp: '27.0'
+            elixir: '1.17'
+            experimental: true
     steps:
     - uses: actions/checkout@v4
     - name: Set up Elixir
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -41,14 +49,22 @@ jobs:
   Test-gun1:
     runs-on: ubuntu-latest
     name: OTP ${{matrix.otp}} / Elixir ${{matrix.elixir}} - Gun1
+    continue-on-error: ${{ matrix.experimental }}
```
nathany-copia, line 52: I'm not sure why there is a separate section for gun? It seems like these tests are also run by the previous test run.

.github/workflows/test.yml
```diff
@@ -83,8 +99,8 @@ jobs:
     - name: Set up Elixir
       uses: erlef/setup-beam@v1
       with:
-        elixir-version: '1.13'
-        otp-version: '24.3'
+        elixir-version: '1.15'
+        otp-version: '25.3'
         version-type: strict
```
nathany-copia, line 104: having an entirely separate job just to check code formatting may be overkill?

unless the idea is to continue-on-error for this particular check, but it's not running with a matrix right now anyway 🤔

.github/workflows/test.yml
yordis: I remember removing this, and @teamon put it back again.

Check the commit history ... I remember I made a mistake so I am extra careful now

.github/workflows/test.yml
yordis: Pick one matrix (add some attribute to it) to do the formatting and lining and whatnot

.github/workflows/test.yml
teamon: It is there to ensure tests pass with gun 1.x too (the default lockfile will use gun 2.x)

There used to be a similar thing for poison 3. 

We could remove it if we drop support for gun 1.x but I’m not sure how big of an issue that would be for tesla users. I think I’d rather keep that CI job than force (or prevent) people to update their deps major versions.

nathany-copia: Thanks for clarifying

nathany-copia:
@yordis @teamon What should be done about all the test failures, even on the "supported" Elixir 1.16 build?

.github/workflows/test.yml
nathany-copia: Is it reasonable to run the gun 1 tests only on an older Elixir / OTP combination instead of the whole matrix? (e.g. OTP 25.3, Elixir 1.15)

.github/workflows/test.yml
nathany-copia: Okay. I can give that a try. It's going to be a little tricky to know if I got that right on the first try though, given the "workflow awaiting approval" for testing CI.

> 1 workflow awaiting approval

nathany-copia:
@yordis @teamon Everything should be ready for a re-run in CI.

I've marked Elixir 1.16 as experimental until those test failures are fixed, which I think could be a separate pull request.

.github/workflows/test.yml
yordis: Yes

.github/workflows/test.yml
nathany-copia: Okay. I ran mix format with a newer version of Elixir (as the old lint job was using 1.13)

nathany-copia:
I think this should be good to merge now.

Then we can follow-up with a PR to get Elixir 1.16 passing and mark it as no longer "experimental" -- if that sounds okay to you?

yordis:
Yeah, that works! Remember `httparrot` is the package that needs to be fixed

nathany-copia:
> Yeah, that works! Remember `httparrot` is the package that needs to be fixed _or replace_

Opened #684 for that topic. Let's see if it can be fixed.

PR #703, opened 2024-08-05
Revert "Revert "feat: add ssl opt to httpc by default""

yordis:
Reverts elixir-tesla/tesla#702


yordis:
I have taken the stand that `Tesla` will no try to reconfigure the defaults of `:httpc` to fix the situation.

I firmly believe that this should be fixed upstream in OTP itself. I can not ensure that I will not break some codebase that may rely on the existing setup.

Thank y'all!

PR #705, opened 2024-08-18
Fix typos and improve language

preciz:
Thank you for reviewing.

### DIFF
--- a/lib/tesla/adapter/mint.ex
+++ b/lib/tesla/adapter/mint.ex
@@ -315,7 +315,8 @@ if Code.ensure_loaded?(Mint.HTTP) do
 
         {:error, _conn, error, _res} ->
           if opts[:close_conn], do: {:ok, _conn} = close(conn)
-          {:error, "Encountered Mint error #{inspect(error)}"}
+          # TODO: (breaking change) fix typo in error message
+          {:error, "Encounter Mint error #{inspect(error)}"}
 
         :unknown ->
           if opts[:close_conn], do: {:ok, _conn} = close(conn)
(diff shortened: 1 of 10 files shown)


Fixed SEARCH/REPLACE block for `lib/tesla/middleware/path_params.ex
yordis, line 1: I do not understand the changes here 🤔

yordis (changes requested):

Fixed SEARCH/REPLACE block for `lib/tesla/middleware/path_params.ex
preciz: Sorry, accidentally committed that. Generated by the aider-chat tool.

lib/tesla/adapter/mint.ex
```diff
@@ -315,7 +315,7 @@ if Code.ensure_loaded?(Mint.HTTP) do
 
         {:error, _conn, error, _res} ->
           if opts[:close_conn], do: {:ok, _conn} = close(conn)
-          {:error, "Encounter Mint error #{inspect(error)}"}
+          {:error, "Encountered Mint error #{inspect(error)}"}
```
yordis, line 318: I am unsure if we should change the error, even when it is a misspelling; I am not sure who would be relying on it ...

preciz: Or maybe not 0 https://github.com/Logflare/logflare/blob/dc9b17541eab4a7118cdc79adaf190b48cecdca3/lib/logflare_web/live/vercellogdrains_live/index.ex#L93

yordis: Do you mind adding a TODO comment to change when we decide to break change?

preciz: Very good idea, done that

yordis (approved):

PR #708, opened 2024-08-27
Make headers addition idempotent

carrascoacd:
## Problem

When used with the `Headers` middleware, `JSON` is sometimes able to add the `content-type` header twice if the former already added it. This prevents the usage of endpoints that have no body (no encodable) but that return JSON. This is not a problem with `hackney`, but if we use `Mint`, the duplicity is not handled and, in consequence, the header is duplicated.

## Proposed solution

Make idempotent the header addition so we allow our middleware to be isolated no matter how it is used. This could be opinionated, since we could rely on clients to design APIs that accept a non nil body, but let's see if you consider this as a good thing to have.

cc/ @yordis

### DIFF
--- a/lib/tesla/middleware/json.ex
+++ b/lib/tesla/middleware/json.ex
@@ -71,6 +71,7 @@ defmodule Tesla.Middleware.JSON do
       {:ok,
        env
        |> Tesla.put_body(body)
+       |> Tesla.delete_header("content-type")
        |> Tesla.put_headers([{"content-type", encode_content_type(opts)}])}
     else
       false -> {:ok, env}
--- a/test/tesla/middleware/json_test.exs
+++ b/test/tesla/middleware/json_test.exs
@@ -117,6 +117,43 @@ defmodule Tesla.Middleware.JsonTest do
     end
   end
 
+  describe "Idempotent content type" do
+    defmodule ContentTypeJSON do
+      @behaviour Tesla.Middleware
+
+      @impl Tesla.Middleware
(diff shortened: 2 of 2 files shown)


yordis:
> When used with the Headers middleware, JSON is sometimes able to add the content-type header twice if the former already added it.

Is this due to a misconfiguration problem?

The HTTP spec allows the headers to be duplicated, and based on the information, it seems that you need to handle adding the middleware when appropriate or fixing the misconfiguration situation.

carrascoacd:
> Is this due to a misconfiguration problem?

Not really, is basically that the JSON middleware adds the header when the body is encodable so it could end up being duplicated. Cowboy fails on duplicated headers under Phoenix so this was a way of preventing it from happening. The responsibility of the middleware overlaps with the Headers middleware as it is adding headers implicitly.

It can be solved in other ways (server side or client side) following the HTTP specs. It is opinionated, so if you see no benefit to it I'll close this.

yordis:
> The responsibility of the middleware overlaps with the Headers middleware as it is adding headers implicitly.

Right, I am saying that you have conflicting configurations; either you remove the `content` from the Header or remove the JSON middleware.

> if you see no benefit to it I'll close this.

I do see the problem and benefits; my concern is that I cannot tell who is OK with such behavior. I must be careful about it.

carrascoacd:
I totally agree, I was not sure either that's why I wanted to know your opinion. Let's close this as it can be solved in different ways.

Thanks dude!

PR #718, opened 2024-10-23
feat: add mox integration test support

yordis:


yordis:
@teamon, I would like to take ownership of introducing some testing tools into Tesla. I have been doing similar work for half a decade by now https://github.com/stueccles/okta-elixir/commits/master/test/support/helpers.ex

I have tested every possible way to set things up, and professionally speaking, I have seen enough patterns by now.
I have opinions about avoiding previous mistakes and making it as professionally ready as possible.

Granted, they are a set of things I would expect people to do. I think that is true of any tool out there. I am OK with it. I will be writing documents and whatnot to explain the situation.

Ultimately, I will provide tools instead of continuing to hurt the professional environment since most people keep making the mistakes I have made many years ago.

mix.exs
```diff
@@ -70,6 +70,7 @@ defmodule Tesla.Mixfile do
       # other
       {:fuse, "~> 2.4", optional: true},
       {:telemetry, "~> 0.4 or ~> 1.0", optional: true},
+      {:mox, "~> 1.0.0", optional: true},
```
aloukissas, line 73: Any reason you're not pointing to latest `mox` (1.2.0)? The previous `mix.lock` was resolving to this, too

guides/howtos/test-using-mox.md
```diff
@@ -0,0 +1,88 @@
...
+define expectations and verify them:
+
+```elixir
+defmodule MyApp.FeatureTest do
+  use ExUnit.Case, async: true
+
+  require Tesla.Test
+
+  setup context, do: Mox.set_mox_from_context(context)
+  setup context, do: Mox.verify_on_exit!(context)
```
teamon, line 43: isn't the "standard" way of doing this something like:


```ex
import Mox

setup :verify_on_exit!
```

? 

I know it's the same, but I think the imported version is what Mox uses in its docs.

guides/howtos/test-using-mox.md
```diff
@@ -0,0 +1,88 @@
...
+
+  test "example test" do
+    # Expect a single HTTP request to be made and return a JSON response
+    Tesla.Test.expect_tesla_call(
```
teamon, line 47: Is it necessary to repeat "tesla" in the function name? 
I think `Tesla.Test.expect_call` or just `Tesla.Test.expect` would do.

guides/howtos/test-using-mox.md
```diff
@@ -0,0 +1,88 @@
...
+    assert :ok = create_user!(%{username: "johndoe"})
+
+    # Verify that the HTTP request was made and matches the expected parameters
+    Tesla.Test.assert_received_tesla_call(env, [])
+    Tesla.Test.assert_tesla_env(env, %Tesla.Env{
+      method: :post,
+      url: "https://acme.com/users",
+      body: %{username: "johndoe"},
+      status: 200,
+    })
```
teamon, line 63: What's the advantage over standard pattern match `assert %Env{...} = ...`?

teamon (commented):
This is definitely a good direction but I'm not 100% sold on the api

guides/howtos/test-using-mox.md
yordis: I am focusing on removing magical things that, although "standard" adds to confusion, especially for newbies;

I expect users to be familiar with `Mox` to some extent, for sure, but also avoid trivial problems I seen before if possible.

guides/howtos/test-using-mox.md
yordis: Usually. I put importance on the functions over the module.

If people prefer `import Tesla.Test` then they will avoid collisions with 
Other assertions are available if people prefer to import the assertions functions. Also, "call" makes it clear what that one is about.

guides/howtos/test-using-mox.md
yordis: Some convenience is based on previous experience of taking the given `env` **as is** and focusing on the matching alone, most of the time without having to unmarshal back the data in the JSON cases and excluding headers.

guides/howtos/test-using-mox.md
teamon: Give the `import Mox` is the very first thing in [mox docs](https://hexdocs.pm/mox/Mox.html) the `setup context ...` syntax was actually more confusing for me (is something special needed for tesla & mox to work together?).

I'd love to get some more opinions on it.

guides/howtos/test-using-mox.md
teamon: I'd say we should either:
- Option A: Keep `expect_tesla_call` and `import Tesla.Test` in all examples
- Option B: Change to `Tesla.Test.expect` and never use imports

guides/howtos/test-using-mox.md
teamon: I could imagine instead of:

```ex
Tesla.Test.assert_received_tesla_call(env, [])
Tesla.Test.assert_tesla_env(env, %Tesla.Env{
  method: :post,
  url: "https://acme.com/users",
  body: %{username: "johndoe"},
  status: 200,
})
```

doing something like this:

```ex
assert_received %Tesla.Env{
  method: :post,
  url: "https://acme.com/users",
  body: body,
  status: 200,
}

assert %{username: "johndoe"} = as_json(body)
```

This way you can use all the standard assertions with all their variants (`(assert|refute)(_receive)(d)`), extra params and test failure printout and it's clear what is really happening and what is being tested.

yordis: That works as well; it is all about styling at that point, to be honest; 

image

guides/howtos/test-using-mox.md
yordis: Doing Option A; is what I had in mind, it is a matter of time 😄

guides/howtos/test-using-mox.md
yordis: fixing the docs; no strong opinions at all

guides/howtos/test-using-mox.md
yordis: done

guides/howtos/test-using-mox.md
yordis: showed the an extra example; we can always improve it as we go

guides/howtos/test-using-mox.md
yordis: done

PR #719, opened 2024-10-23
feat: release-please and conventional commit

doomspork:
@yordis I think this should be a good start to what you're looking for if I understood you correctly.

You may want to config .github/release-please-config.json like we've done in BEAM Community to specify changelog-sections


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,21 @@
...
+    steps:
+      - id: release
+        name: Release
+        uses: google-github-actions/release-please-action@v4
+        with:
+          command: manifest
+          config-file: .github/release-please-config.json
+          manifest-file: .github/release-please-manifest.json
+          release-type: elixir
+          token: ${{ secrets.GH_PERSONAL_ACCESS_TOKEN }}
```
doomspork, line 21: I _think_ you're going to need a PAT for this to work 🤔

yordis: @teamon, I have been trying to contact you on Slack and X. What would be a good way to reach out?

PR #723, opened 2024-10-25
chore: fixing example and removing reminder use Tesla showcasing

yordis:


arcanemachine:
Hey @yordis, sorry if this is already posted somewhere else, but is the `use Tesla` syntax deprecated for creating new Clients?

I do not see the `use Tesla` syntax anywhere in the docs and it looks like it has been removes in favor of using `Tesla.client/2` to define Clients.

Just seeking some clarification on the situation here. Thanks.

yordis:
Hey, @arcanemachine, please subscribe to https://github.com/elixir-tesla/tesla/issues/729. I made an oversight, and I need to find time to correct it.

PR #725, opened 2024-10-25
Add assert_receive_tesla_call/3 test macro

visciang:
This PR extends the `Tesla.Test` utility with the `assert_receive_tesla_call/3` macro, in line with the existing 
 `assert_received_tesla_call/3`.

@yordis I haven’t added coverage for this macro in `test_test.exs`. Feel free to adjust or refactor the tests to fit a more idiomatic style if needed.

### DIFF
--- a/lib/tesla/test.ex
+++ b/lib/tesla/test.ex
@@ -174,11 +174,36 @@ defmodule Tesla.Test do
       end
   """
   defmacro assert_received_tesla_call(expected_env, expected_opts \\ [], opts \\ []) do
-    adapter = fetch_adapter!(opts)
+    opts = Keyword.put(opts, :timeout, 0)
 
     quote do
-      assert_received {Tesla.Test,
-                       {unquote(adapter), :call, [unquote(expected_env), unquote(expected_opts)]}}
+      unquote(__MODULE__).assert_receive_tesla_call(
+        unquote(expected_env),
+        unquote(expected_opts),
+        unquote(opts)
+      )
+    end
+  end
+
+  @doc """
+  Asserts that the current process received or is going to receive a `TeslaMox` message.
+  It uses `assert_receive/1` under the hood.
+
+  ## Parameters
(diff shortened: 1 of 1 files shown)


lib/tesla/test.ex
```diff
@@ -174,11 +174,36 @@ defmodule Tesla.Test do
       end
   """
   defmacro assert_received_tesla_call(expected_env, expected_opts \\ [], opts \\ []) do
+    opts = Keyword.put(opts, :timeout, 0)
+
+    quote do
+      unquote(__MODULE__).assert_receive_tesla_call(
```
yordis, line 180: I thought the intent was to add `assert_receive` along with `assert_received`; this will go thru the same code path of `receive` (without the `d`)

visciang: > I could understand that maybe (I need to check) under the hood `assert_receive` and `assert_received` could be the same, just timeout 0; but I rather keep it different here as well instead of sharing the code path

Exactly, `assert_received` is defined in terms of `assert_receive` with a 0 timeout:
https://github.com/elixir-lang/elixir/blob/90f504e9c389b1c6a36580b82a815621735d670a/lib/ex_unit/lib/ex_unit/assertions.ex#L496

That’s why I thought it might make sense to follow the standard Elixir library’s approach here.

But honestly, as long as we get the behavior we need—and can tweak the timeout in `assert_receive_tesla_call`—feel free to refactor it in whatever way feels most natural with the current Tesla.Test setup!

yordis: I get it, A long time ago, I looked down there; still, that is true today, and I would rather use their built-in functions.

Are you able to fix the situation and add some unit test? Otherwise, it will no get in any time soon, I am dealing with plenty of things right now, so I need your help

yordis: Feel free to reopen and follow up on the feedback

PR #744, opened 2025-01-13
Use rescue to avoid dependency inversion

carrascoacd:
## Problem

When using Finch is expected to receive a RuntimeError if the connection pool is exhausted, see https://github.com/sneako/finch/blob/main/test/finch_test.exs#L561

This is a problem because in a service that uses a lower tier API than the service itself, a dependency inversion of tiers can happen. For example, a tier 2 service calling a tier 3 service. 

The tier inversion happens because we are not rescuing the error and returning it into a tuple, so the operation can work without problems under this situation.

## Proposed solution

Encapsulate the error in the error tuple. I'm adding this solution here, but I want to know if you consider that it would be best suited in the Finch code itself. Changing `reraise` basically.

cc/ @yordis

### DIFF
--- a/lib/tesla/adapter/finch.ex
+++ b/lib/tesla/adapter/finch.ex
@@ -65,15 +65,20 @@ if Code.ensure_loaded?(Finch) do
       req_opts = Keyword.take(opts, [:pool_timeout, :receive_timeout])
       req = build(env.method, url, env.headers, env.body)
 
-      case request(req, name, req_opts, opts) do
-        {:ok, %Finch.Response{status: status, headers: headers, body: body}} ->
-          {:ok, %Tesla.Env{env | status: status, headers: headers, body: body}}
+      try do
+        case request(req, name, req_opts, opts) do
+          {:ok, %Finch.Response{status: status, headers: headers, body: body}} ->
+            {:ok, %Tesla.Env{env | status: status, headers: headers, body: body}}
 
         {:error, %Mint.TransportError{reason: reason}} ->
           {:error, reason}
 
-        {:error, reason} ->
-          {:error, reason}
+          {:error, reason} ->
+            {:error, reason}
+        end
+      rescue
+        e in RuntimeError ->
+          {:error, {:runtime_error, e}}
       end
     end
 
--- a/lib/tesla/adapter/finch.ex
+++ b/lib/tesla/adapter/finch.ex
@@ -70,8 +70,8 @@ if Code.ensure_loaded?(Finch) do
           {:ok, %Finch.Response{status: status, headers: headers, body: body}} ->
             {:ok, %Tesla.Env{env | status: status, headers: headers, body: body}}
 
-        {:error, %Mint.TransportError{reason: reason}} ->
-          {:error, reason}
+          {:error, %Mint.TransportError{reason: reason}} ->
+            {:error, reason}
 
           {:error, reason} ->
             {:error, reason}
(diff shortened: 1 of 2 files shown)


yordis (commented):
I am concerned with this one; it would change the public API, I am not saying it is bad, but I must be careful here.

carrascoacd:
What do you mean with changing the public API? As far as I know, the adapter is still returning the error tuple as usual but with a new reason.

Could you provide me with an example?

yordis:
```elixir
{:error, {:runtime_error, e}}
```

That is a new pattern matching that must be taken into consideration. I just do not know what assumptions people made, which leads me to:

```elixir
e in RuntimeError ->
```

What would happen if somebody was expecting such `RuntimeError` exception btw?

carrascoacd:
Regarding the pattern matching, we could change it to a raw string.

> What would happen if somebody was expecting such RuntimeError exception btw?

Good question. In my opinion, returning a RuntimeError is breaking the contract the Adapter defines, that us basically a ok or error tuple. This is already breaking the contract in fact. If you use Mint adapter you should be able to replace it with Finch adapter without changing your codebase. This is basically the typical substitution principle.

But I get your point, that's what I'm no sure if this should be added in the Finch library instead. The fact is that, this adapter is not valid for production usage as it is (without catching the error).

Let me know if you prefer to close this and move this PR yo Finch instead.

Thanks!

carrascoacd:
@yordis contributors from Finch shared their point of view. Should we make the change here or what do you propose? What about making this change under parameter so we ensure backwards compatibility?

Let me know your thoughts! Thanks

yordis:
As @sneako said,

>  I don't think we should make this breaking change at this time.

I can not change the semantics because I don't know who depends on that behavior. I already broke people's apps by changing the semantics in `:httpc`, so I am a bit more conservative with such a change.

> What about making this change under parameter so we ensure backwards compatibility?

I would suggest to create a middleware, you can make that middleware aware of a given option to change the behaviour.

carrascoacd:
> I would suggest to create a middleware, you can make that middleware aware of a given option to change the behaviour

Cool! Just to double check. Do you suggest to create a different and new middelware or extend the current one with a new configuration option @yordis ?

yordis:
I was thinking of a new middleware in your app.

What current middleware are you referring to? Tesla does not comes with any built-in middleware setup, intentionally.

carrascoacd:
Oh, sorry @yordis. I mixed the term Adapter with the term Middleware haha. I was talking about extending the current Finch Adapter with an option. Something like `raise: true` by default, so it can be customized.

It will be much easier for your clients to change a system through the global config options rather than changing every client to incorporate an extra middeware.

yordis:
Hey @carrascoacd thank you so much for taking the time to contribute, I really appreciate it.

As @sneako said:

> I don't think we should make this breaking change at this time.

I can not make this breaking change in Tesla either, and I do not know who relies on proper exception handling.

The good thing is that you can always create a `SafeMiddleware` or something similar that tries to catch all exceptions.

Once again, thank you so much!

carrascoacd:
Hi @yordis you are welcome.

> I can not make this breaking change in Tesla either, and I do not know who relies on proper exception handling.

I think I didn't explain myself. I'm not suggesting a breaking change but a new feature in the adapter that allows customizing the behaviour, hence a minor change. For example:

```elixir
Keyword.take(opts, [:pool_timeout, :receive_timeout, :raise_errors])

# if not given, keep the current behaviour
```

Let me know if it makes sense and I'll submit a new change.

yordis:
I agree with the intent; for what it is worth, we need to normalize some "reserved" options, such as timeout and returns (tuple-result, exceptions) across everything.

That said, we are responsible for every new flag and need to consider it across all adapters, not just this one.

Before getting to implementation detail, we should discuss how will be done and again, think about multiple adapters.

PR #746, opened 2025-01-24
Fix. Mock not working for grandchildren and beyond

carrascoacd:
## Context

`!is_nil(Process.get(ancestor, __MODULE__))` always returns true, hence the `find` fn returns the first ancestor. 

## Solution

Iterate properly over ancestors trying to find the `__MODULE__` key.

cc/ @yordis

### DIFF
--- a/lib/tesla/mock.ex
+++ b/lib/tesla/mock.ex
@@ -231,7 +231,13 @@ defmodule Tesla.Mock do
   defp pdict_get do
     pid_holder =
       Enum.find(Process.get(:"$ancestors", []), self(), fn ancestor ->
-        !is_nil(Process.get(ancestor, __MODULE__))
+        is_holder? =
+          ancestor
+          |> Process.info()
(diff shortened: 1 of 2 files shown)


yordis (approved):
🚀

tomekowal:
Hey! Thanks for that fix. I was bitten by the same issue recently.
However, I am wondering if the code works as intended.
E.g. it starts searching for the mock from the "oldest" or "most root" ancestor.
If, for some reason, parent process defined a mock, then defining new one in the test process would not work.
Also, I don't think `$ancestors` is the correct thing to use.
E.g. for supervised tasks, the ancestor is the task supervisor and not the test process.

I believe, something like this should work better:

```
pids = [self() | Enum.reverse(Process.get(:"$callers", []))]

pid_holder = Enum.find(pids, nil, fn pid -> ... end)
```

If you agree with my reasoning, I am happy to make a PR :)

Issue #743, opened 2025-01-02
Default `Tesla.Middleware.JSON` engine to `JSON` module for Elixir 1.18


Wigny:
Currently `Tesla.Middleware.JSON` defaults to `Jason` as its JSON decoder and encoder, but since Elixir 1.18 now there is a built-in alternative that could be used as the default engine when running in Elixit 1.18, eliminating a dependency on another library.

Right now, to use the new module, I believe we could use the `encode` and `decode` options, but it cannot be directly passed in the `engine` option since `JSON` does not have an `encode/1` function and also do not support `engine_opts`.

```elixir
{Tesla.Middleware.JSON, encode: &{:ok, JSON.encode!(&1)}, decode: &JSON.decode/1 }
```

adamu:
Although the docs say that only `Jason`, `JSX`, and `Poison` are supported, in fact the code does not verify this and accepts any module as the `:engine`, assuming there are `encode/2` and `decode/2` functions that return ok/error tuples.

Adding support for `JSON` without making breaking the ability to supply other `:engines` might be tricky.

yordis:
Due to backward compatibility, I will keep Jason as default; once Elixir 1.18+ is required, I can make the change.

For now, users can opt-in.

PR #747, opened 2025-01-26
Tesla.Middleware.JSON: Add support for Elixir 1.18's JSON module

adamu:
This is a proposal for one way to add support to the built-in `JSON` module available in Elixir 1.18.

There were a few challenges with this, because Tesla makes assumptions about how an engine should behave that JSON doesn't conform to. To work around this I created a `JSONAdapter` module that conforms to Tesla's expectations, and allows users to simply specify

    engine: JSON

for common cases, which will give people an easy path to migrate away from `Jason`, for example.

Elixir's built-in JSON module also supports custom encoders and decoders, which could theoretically be supplied via this middlewear's `engine_opts` param. But considering that would require additional wrapper/adapter logic, and the same thing can be achieved using the `encoder:` and `decoder:` options, I decided that might be overkill.

Long term, I don't think the `:engine` option makes much sense, because there are too many assumptions involved without defining a proper behaviour. Perhaps deprecating it and focussing on the `:encoder` and `:decoder` options might make sense?

## Docs preview

image

## Related issues
* #743

### DIFF
--- a/lib/tesla/middleware/json.ex
+++ b/lib/tesla/middleware/json.ex
@@ -12,6 +12,15 @@ defmodule Tesla.Middleware.JSON do
   mix deps.compile tesla
   ```
 
+  > #### Using built-in `JSON` from Elixir 1.18 {: .info}
+  >
+  > This middleware supports the built-in `JSON` module introduced in ELixir 1.18, but for historical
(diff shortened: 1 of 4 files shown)


yordis:
🚀 💜

PR #750, opened 2025-02-03
Fix mocks for supervised tasks

tomekowal:
This PR is a follow up of this comment:
https://github.com/elixir-tesla/tesla/pull/746#issuecomment-2629558637

### DIFF
--- a/lib/tesla/mock.ex
+++ b/lib/tesla/mock.ex
@@ -229,23 +229,12 @@ defmodule Tesla.Mock do
 
   # Gets the mock fun for the current process or its ancestors
   defp pdict_get do
-    pid_holder =
-      Enum.find(Process.get(:"$ancestors", []), self(), fn ancestor ->
-        is_holder? =
-          ancestor
-          |> Process.info()
-          |> Keyword.get(:dictionary)
-          |> Keyword.get(__MODULE__)
+    potential_mock_holder_pids = [self() | Enum.reverse(Process.get(:"$callers", []))]
 
-        !is_nil(is_holder?)
-      end)
-      |> case do
-        nil -> raise "Unknown pid_holder in mock"
-        pid when is_pid(pid) -> pid
-        name when is_atom(name) -> Process.whereis(name)
-      end
-
-    pid_holder |> Process.info() |> Keyword.get(:dictionary) |> Keyword.get(__MODULE__)
+    Enum.find_value(potential_mock_holder_pids, nil, fn pid ->
+      {:dictionary, process_dictionary} = Process.info(pid, :dictionary)
+      Keyword.get(process_dictionary, __MODULE__)
+    end)
   end
 
   defp agent_set(fun) do
--- a/test/tesla/mock_test.exs
+++ b/test/tesla/mock_test.exs
@@ -113,6 +113,41 @@ defmodule Tesla.MockTest do
     end
   end
 
+  describe "supervised task" do
+    test "allows mocking in the caller process" do
+      # in real apps, task supervisor will be part of the supervision tree
+      # and it won't be an ancestor of the test process
+      # to simulate that, we will set the mock in a task
+      #
+      # test_process
+      # |-mocking_task will set the mock and create the supervised task
+      # `-task supervisor
+      #   `- supervised_task
+      # this way, mocking_task is not an $ancestor of the supervised_task
+      # but it is $caller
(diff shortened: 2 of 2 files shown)


lib/tesla/mock.ex
```diff
@@ -229,23 +229,12 @@ defmodule Tesla.Mock do
...
-      |> case do
-        nil -> raise "Unknown pid_holder in mock"
-        pid when is_pid(pid) -> pid
-        name when is_atom(name) -> Process.whereis(name)
-      end
+    potential_mock_holder_pids = [self() | Enum.reverse(Process.get(:"$callers", []))]
 
-    pid_holder |> Process.info() |> Keyword.get(:dictionary) |> Keyword.get(__MODULE__)
+    Enum.find_value(potential_mock_holder_pids, nil, fn pid ->
+      {:dictionary, process_dictionary} = Process.info(pid, :dictionary)
```
tomekowal, line 235: Using `Process.info(pid, :dictionary)` instead of just `Process.info(pid)` reduces the amount data copied between processes.

lib/tesla/mock.ex
```diff
@@ -229,23 +229,12 @@ defmodule Tesla.Mock do
...
-      end)
-      |> case do
-        nil -> raise "Unknown pid_holder in mock"
-        pid when is_pid(pid) -> pid
-        name when is_atom(name) -> Process.whereis(name)
-      end
+    potential_mock_holder_pids = [self() | Enum.reverse(Process.get(:"$callers", []))]
 
-    pid_holder |> Process.info() |> Keyword.get(:dictionary) |> Keyword.get(__MODULE__)
+    Enum.find_value(potential_mock_holder_pids, nil, fn pid ->
```
tomekowal, line 234: Using `find_value` instead of `find` returns the mock directly, so we don't have to call `Process.info` for a second time after we got the pid.

lib/tesla/mock.ex
```diff
@@ -229,23 +229,12 @@ defmodule Tesla.Mock do
...
-          |> Keyword.get(__MODULE__)
-
-        !is_nil(is_holder?)
-      end)
-      |> case do
-        nil -> raise "Unknown pid_holder in mock"
-        pid when is_pid(pid) -> pid
-        name when is_atom(name) -> Process.whereis(name)
-      end
+    potential_mock_holder_pids = [self() | Enum.reverse(Process.get(:"$callers", []))]
```
tomekowal, line 232: We search for mocks hierarchically. Even though, it is unlikely that we've set mocks both in the caller and in the process, I would be super if I used mock in the current process and it found a mock from somewhere else.

Hence, I start searching from mocks from the current process and then up the callers hierarchy.

In most situations `$callers` and `$ancestors` are the same. E.g. spawning a task, starting a GenServer puts the same pid in both.

The notable difference is starting a supervised task because the ancestors of a supervised task is its supervisor, while the caller is the process that asked to spawn it.

I believe that for mocking `$callers` is the correct one to use.

lib/tesla/mock.ex
```diff
@@ -229,23 +229,12 @@ defmodule Tesla.Mock do
...
-      Enum.find(Process.get(:"$ancestors", []), self(), fn ancestor ->
-        is_holder? =
-          ancestor
-          |> Process.info()
-          |> Keyword.get(:dictionary)
-          |> Keyword.get(__MODULE__)
-
-        !is_nil(is_holder?)
-      end)
-      |> case do
```
tomekowal, line 242: That `case` statement is a leftover from the previous solution that assumed using non-existent `Process.get(pid, value)`. Now, it is not necessary.

carrascoacd:
It makes sense to me, thanks for the improvements!

yordis:
btw, are you familiar with https://hexdocs.pm/tesla/Tesla.Test.html and https://hexdocs.pm/tesla/1-testing.html ideally you would migrate to use the module

yordis:
Thank you @tomekowal 🚀

tomekowal:
Thanks @yordis I know Tesla.Mock is getting deprecated. We are in the process of upgrading from Tesla 1.4 to 1.13, so we have four years of work to migrate :) `Tesla.Mock` worked flawlessly so far :)
Again, kudos for keeping the library up and running all this years!

yordis:
I dont mind keeping it around for a long time, so don't sweat over it.

Let me know if you need any help with the migration! I wrote a guide for it https://hexdocs.pm/tesla/v1-macro-migration.html

Issue #753, opened 2025-02-22
Incompatibility with hackney 1.22.0


halostatue:
```
The following arguments were given to Tesla.Adapter.Hackney.handle/2:

         # 1
         {:connect_error, {:error, :econnrefused}}

         # 2
         [recv_timeout: 150000]

     Attempted function clauses (showing 5 out of 5):

         defp handle({:error, _} = error, _opts)
         defp handle({:ok, status, headers}, _opts)
         defp handle({:ok, ref}, _opts) when is_reference(ref)
         defp handle({:ok, status, headers, ref}, opts) when is_reference(ref)
         defp handle({:ok, status, headers, body}, _opts)
```

yordis:
@halostatue are you able to create a PR for it?

I know people are rushing to upgrade hackney but I am away from the computer.

We need to pattern match on that `{:connect_error, {:error, reason}}`

```elixir
defp handle({:connect_error, {:error, reason}}, _opts), do: {:error, reason}
```

PR #754, opened 2025-02-22
fix: add basic Hackney 1.22 support: {:connect_error, _}

halostatue:
Fixes: #753


halostatue:
I did look to see how to add a test, but it doesn't look like it's a supported case for httparrot.

PR #760, opened 2025-03-12
feat: add url policy to logger middleware

yordis:


yordis:
@teamon I am still unsure if this idea should be custom-formatted in the app instead of extending the logger's behaviour.

Primarily because the configuration only makes sense for the default format AND requires proper middleware setup.

I am inclined to recommend that people create their format after thinking about it 🤔 

I suggest providing another formatting function that people could opt for in a similar situation.

yordis:
I talked myself out, custom formatter should be the solution here

PR #763, opened 2025-04-07
fix: move regexes out of module attributes to fix compatibility with OTP 28

TBK145:


yordis:
Hey there, thank you so much for the help. Do you mind sharing more details about the OTP 28 changes and Elixir support for it?

TBK145:
Hi, OTP 28 changes the regex implementation from PCRE2 to RE2. As far as I can tell, most things stay the same, but compilation of a regex now returns a reference, which cannot be escaped and therefore not used as a module attribute.
Newer versions of Elixir will have a workaround to still make it compile, but also start emitting deprecation warnings for this: https://github.com/elixir-lang/elixir/pull/14381

yordis:
Thank you for the explanation and reference 💜 🚀

Issue #765, opened 2025-04-09
Mock no longer works on spawned processes


yastanotheruser:
Since version 1.14.0, `Tesla.Mock.mock` has stopped working for processes spawned by a test or a callback. I think this is a regression due to #750.

### Version 1.13

```elixir
# tesla_1.13_mock.exs
Mix.install([
  {:jason, "~> 1.4"},
  {:tesla, "~> 1.13.0"}
])

Application.ensure_all_started(:tesla)
Application.put_env(:tesla, :adapter, Tesla.Mock)

defmodule MyAgent do
  use Agent

  def start_link(_arg) do
    Agent.start_link(fn -> Tesla.get!("https://httpbin.org/get") end, name: __MODULE__)
  end
end

Tesla.Mock.mock(fn _ -> Tesla.Mock.json(%{"url" => "https://httpbin.org/get"}) end)
{:ok, _pid} = MyAgent.start_link([])
IO.inspect(:sys.get_state(MyAgent), label: "state")
```

```console
$ elixir tesla_1.13_mock.exs
state: %Tesla.Env{
  method: nil,
  url: "",
  query: [],
  headers: [{"content-type", "application/json"}],
  body: "{\"url\":\"https://httpbin.org/get\"}",
  status: 200,
  opts: [],
  __module__: nil,
  __client__: nil
}
```

### Version 1.14

```elixir
# tesla_1.14_mock.exs
Mix.install([
  {:jason, "~> 1.4"},
  {:tesla, "~> 1.14.0"}
])

Application.ensure_all_started(:tesla)
Application.put_env(:tesla, :adapter, Tesla.Mock)

defmodule MyAgent do
  use Agent

  def start_link(_arg) do
    Agent.start_link(fn -> Tesla.get!("https://httpbin.org/get") end, name: __MODULE__)
  end
end

Tesla.Mock.mock(fn _ -> Tesla.Mock.json(%{"url" => "https://httpbin.org/get"}) end)
{:ok, _pid} = MyAgent.start_link([])
IO.inspect(:sys.get_state(MyAgent), label: "state")
```

```console
$ elixir tesla_1.14_mock.exs
** (EXIT from #PID<0.99.0>) an exception was raised:
    ** (Tesla.Mock.Error) There is no mock set for process #PID<0.9.0>.
Use Tesla.Mock.mock/1 to mock HTTP requests.

See https://github.com/teamon/tesla#testing

[...]

PR #771, opened 2025-05-02
Bring back searching for mocks in ancestors

tomekowal:
Hi, @yordis That should fix the #765

As you suggested, I added a TODO with explanation that checking $ancestors is somewhat hacky (both in the mock code and in the test for it.
The example in the test is almost 1:1 the example from the issue.


tomekowal:
Always happy to help! :)

PR #778, opened 2025-06-02
fix: :arrow_up: Bump up hackney as audit fails for version < 1.24.0

yatender-oktalk:
https://github.com/elixir-tesla/tesla/issues/777


yordis (changes requested):
Hey @yatender-oktalk thank you so much for the PR; one more thing, could you change the SemVer from `fix` to `chore`

As I said above, hackney is an optional deps so we do not need to release a new version.

yatender-oktalk:
> I just noticed that upgrades the pkg #775

Nice!

PR #779, opened 2025-06-05
feat: add logging options to use `:level` instead of deprecated `:log_level`

### LINKED ISSUE #379: Documentation Issues: Logger Middleware

yordis:
closes #379

### DIFF
--- a/guides/explanations/3.adapter.md
+++ b/guides/explanations/3.adapter.md
@@ -14,7 +14,7 @@ defmodule Tesla.Adapter.Req do
   @behaviour Tesla.Adapter
 
   @impl Tesla.Adapter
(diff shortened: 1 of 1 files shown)


yordis:
@teamon please share a review over here

teamon (approved):
LGTM! <3 <3 <3

PR #800, opened 2025-08-04
docs: add Finch adapter specific stream option

victor23k:
The docs are missing one specific option for the Finch adapter: `:response`. It is mentioned in the [cheat sheet docs](https://hexdocs.pm/tesla/general.html#streaming-response-body) but not in the Finch Adapter.


yordis (approved):

yordis:
🚀 💜

PR #813, opened 2025-11-25
fix: Handle breaking change in Finch.stream API

howleysv:
Finch made a breaking change to the function signature for `Finch.stream/5` [here](https://github.com/sneako/finch/pull/295) which was shipped in version `0.20.0`. This PR does conditional compilation of the response handler so that dialyzer is happy no matter which version of finch you're using.

### DIFF
--- a/lib/tesla/adapter/finch.ex
+++ b/lib/tesla/adapter/finch.ex
@@ -123,10 +123,9 @@ if Code.ensure_loaded?(Finch) do
 
       task =
         Task.async(fn ->
-          case Finch.stream(req, name, nil, fun, opts) do
-            {:ok, _acc} -> send(owner, {ref, :eof})
-            {:error, error} -> send(owner, {ref, {:error, error}})
-          end
+          req
+          |> Finch.stream(name, nil, fun, opts)
+          |> handle_stream_response(ref, owner)
         end)
 
       receive do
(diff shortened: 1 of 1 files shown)


lib/tesla/adapter/finch.ex
```diff
@@ -153,5 +152,19 @@ if Code.ensure_loaded?(Finch) do
...
+    end
+
+    if Application.spec(:finch, :vsn) >= ~c"0.20.0" do
+      defp handle_stream_response({:error, error, _acc}, ref, owner) do
+        send(owner, {ref, {:error, error}})
+      end
+    else
+      defp handle_stream_response({:error, error}, ref, owner) do
+        send(owner, {ref, {:error, error}})
+      end
```
yordis: coukd we pattern match in both returns without checking `finch` 🤔 so which ever one matches that is the return type of thing

yordis: aaahh yeah yeah 😭  what do you think? I don't mind the explicit ignore 🤔 

Anyway, no strong preferences, except the safest one

lib/tesla/adapter/finch.ex
```diff
@@ -153,5 +152,19 @@ if Code.ensure_loaded?(Finch) do
           {:error, :timeout}
       end
     end
+
+    defp handle_stream_response({:ok, _acc}, ref, owner) do
+      send(owner, {ref, :eof})
+    end
+
+    if Application.spec(:finch, :vsn) >= ~c"0.20.0" do
```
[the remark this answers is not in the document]
howleysv: oh nice catch, and TIL about the `Version` module in the stdlib that seems ideal for this. I'll whip up a fix

PR #878, opened 2026-05-21
feat(form-urlencoded): support nested bodies via :brackets option

tsitsiridakisiosif:
- Native `URI.encode_query/1` rejects nested maps and lists, which forces callers integrating with APIs like Stripe to hand-roll bracket-notation encoders or pull in `Plug.Conn.Query` just for the form-encoded case.
- Aligns Tesla with the de-facto convention used by Stripe's official SDKs (node/python/ruby) and PHP's `http_build_query` so users can pass nested payloads directly without a custom encoder.
- Opt-in via `encode: :brackets`; default behavior is unchanged.

### DIFF
--- a/lib/tesla/middleware/form_urlencoded.ex
+++ b/lib/tesla/middleware/form_urlencoded.ex
@@ -225,9 +225,11 @@ defmodule Tesla.Middleware.FormUrlencoded do
 
   defp encode_deep_object(data) do
     data
-    |> Enum.flat_map(fn {key, value} -> encode_value(value, [key]) end)
+    |> Enum.flat_map(&encode_root_entry/1)
     |> Enum.join("&")
   end
+
+  defp encode_root_entry({key, value}), do: encode_value(value, [key])
 
   defp encode_value(nil, _path), do: []
 
@@ -240,23 +242,17 @@ defmodule Tesla.Middleware.FormUrlencoded do
   end
 
   defp encode_value(value, path) when is_map(value) do
-    Enum.flat_map(value, fn {key, nested_value} ->
-      encode_value(nested_value, [key | path])
-    end)
+    Enum.flat_map(value, &encode_keyed_entry(&1, path))
   end
 
   defp encode_value(value, path) when is_list(value) do
     if Keyword.keyword?(value) do
-      Enum.flat_map(value, fn {key, nested_value} ->
-        encode_value(nested_value, [key | path])
-      end)
+      Enum.flat_map(value, &encode_keyed_entry(&1, path))
     else
       value
       |> Enum.reject(&is_nil/1)
       |> Enum.with_index()
-      |> Enum.flat_map(fn {nested_value, index} ->
-        encode_value(nested_value, [index | path])
-      end)
+      |> Enum.flat_map(&encode_indexed_entry(&1, path))
     end
   end
 
@@ -264,13 +260,17 @@ defmodule Tesla.Middleware.FormUrlencoded do
     ["#{encode_path(path)}=#{encode_part(value)}"]
   end
 
+  defp encode_keyed_entry({key, value}, path), do: encode_value(value, [key | path])
+
+  defp encode_indexed_entry({value, index}, path), do: encode_value(value, [index | path])
+
   defp encode_path(path) do
     [root | rest] = Enum.reverse(path)
 
-    Enum.reduce(rest, encode_part(root), fn part, encoded ->
-      "#{encoded}[#{encode_part(part)}]"
-    end)
+    Enum.reduce(rest, encode_part(root), &append_bracket/2)
   end
+
+  defp append_bracket(part, encoded), do: "#{encoded}[#{encode_part(part)}]"
 
   defp encode_part(value) when is_atom(value),
     do: value |> Atom.to_string() |> URI.encode_www_form()
--- a/lib/tesla/middleware/form_urlencoded.ex
+++ b/lib/tesla/middleware/form_urlencoded.ex
@@ -59,75 +59,31 @@ defmodule Tesla.Middleware.FormUrlencoded do
   Myclient.post(client, "/url", %{key: %{nested: "value"}})
   ```
 
-  ## Serialization Styles
+  ## `encode: :deep_object`
 
-  The `:encode` option supports built-in serialization styles that mirror
-  OpenAPI's Encoding Object `style` field. OpenAPI defines four values:
-  `form` (default), `spaceDelimited`, `pipeDelimited`, and `deepObject`.
-  The middleware currently implements `:deep_object`; the other styles
-  may be added later.
-
-  ### `encode: :deep_object`
-
-  Recursive bracket-notation encoder for bodies that contain maps,
-  structs, or lists. The flat default behavior is unchanged when `:encode`
-  is not set.
-
-  Output shape:
-
-  - Nested maps and structs: `parent[child]=value`.
-  - Lists: `parent[0]=a&parent[1]=b` (numeric indices).
-  - Lists of objects: `items[0][name]=a&items[1][name]=b`.
-
-  OpenAPI defines `deepObject` for object values only and leaves array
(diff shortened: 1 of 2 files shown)


lib/tesla/middleware/form_urlencoded.ex
```diff
@@ -30,13 +30,17 @@ defmodule Tesla.Middleware.FormUrlencoded do
 
   - `:decode` - decoding function, defaults to `URI.decode_query/1`
   - `:encode` - encoding function, defaults to `URI.encode_query/1`
+  - `:deep_object` - when `true`, encode bodies that contain maps, structs,
```
yordis: Could you expand upon the reason behind the usage of boolean here?

yordis (changes requested):

lib/tesla/middleware/form_urlencoded.ex
tsitsiridakisiosif: no reason at all, fixed it

lib/tesla/middleware/form_urlencoded.ex
```diff
@@ -121,14 +190,81 @@ defmodule Tesla.Middleware.FormUrlencoded do
...
   defp do_encode(data, opts) do
-    encoder = Keyword.get(opts, :encode, &URI.encode_query/1)
-    encoder.(data)
+    encoder = Keyword.get(opts, :encode)
+    style = opts |> Keyword.get(:style) |> validate_style!()
+
+    if encoder && style do
+      raise ArgumentError,
+            ":encode and :style options are mutually exclusive; pass one or the other"
+    end
```
yordis: this sounds like a smell to me, should `:encode` be the `:style` as well?

yordis (changes requested):

enilsen16 (commented):
Overall this is a solid, well-tested implementation. I've left inline notes on the specific issues worth addressing before merge — the tuple gap and nil-renumbering footgun are the most actionable.

lib/tesla/middleware/form_urlencoded.ex
```diff
@@ -121,14 +154,97 @@ defmodule Tesla.Middleware.FormUrlencoded do
...
+    raise ArgumentError,
+          "cannot encode #{inspect(module)} struct with :deep_object; " <>
+            "convert it to a map, string, or other primitive before passing it as the body"
+  end
+
+  defp encode_value(value, path) when is_map(value) do
+    Enum.flat_map(value, &encode_keyed_entry(&1, path))
+  end
+
+  defp encode_value(value, path) when is_list(value) do
```
enilsen16: `Keyword.keyword?/1` does a full linear scan of the list to verify every element is a `{atom, _}` tuple. For typical form payloads this is fine, but users passing large integer lists pay the full O(n) check before reaching the indexed branch. Worth a comment so readers know this is intentional.

lib/tesla/middleware/form_urlencoded.ex
```diff
@@ -121,14 +154,97 @@ defmodule Tesla.Middleware.FormUrlencoded do
...
+    if Keyword.keyword?(value) do
+      Enum.flat_map(value, &encode_keyed_entry(&1, path))
+    else
+      value
```
enilsen16: Nil filtering before `with_index` shifts all subsequent indices — `[1, nil, 2]` encodes as `ids[0]=1&ids[1]=2`, not `ids[0]=1&ids[2]=2`. The docs mention this, but it's a silent footgun for APIs that rely on stable positional indexing (e.g. JSON Patch sequences, Stripe expand arrays). Consider either raising on nil-in-list or surfacing a stronger warning in the module doc.

lib/tesla/middleware/form_urlencoded.ex
```diff
@@ -121,14 +154,97 @@ defmodule Tesla.Middleware.FormUrlencoded do
...
+  end
+
+  defp append_bracket(part, encoded) do
+    "#{encoded}[#{encode_part(part)}]"
+  end
+
+  defp encode_part(value) when is_atom(value) do
+    value |> Atom.to_string() |> URI.encode_www_form()
+  end
+
```
enilsen16, line 381: This catch-all calls `to_string/1`, which raises `Protocol.UndefinedError` for tuples (`{1, 10}`) rather than the friendly `ArgumentError` established for structs. Users won't get a helpful message. Consider adding an explicit guard:

```elixir
defp encode_part(value) when is_tuple(value) do
  raise ArgumentError,
        "cannot encode tuple #{inspect(value)} with :deep_object; convert to a list or string first"
end
```

test/tesla/middleware/form_urlencoded_test.exs
```diff
@@ -124,6 +124,154 @@ defmodule Tesla.Middleware.FormUrlencodedTest do
     assert env.body == "decodedbody"
   end
 
+  describe "encode: :deep_object end-to-end through middleware" do
+    defmodule NestedClient do
```
enilsen16: Defining `defmodule NestedClient` inside a `describe` block works but generates a module name that embeds the describe string (e.g. `Tesla.Middleware.FormUrlencodedTest."encode: :deep_object end-to-end through middleware".NestedClient`). Minor cosmetic issue — consider hoisting it to module scope alongside the other client modules in this file.

test/tesla/middleware/form_urlencoded_test.exs
```diff
@@ -124,6 +124,154 @@ defmodule Tesla.Middleware.FormUrlencodedTest do
...
+
+    test "raises clear error for unknown encoder option" do
+      assert_raise ArgumentError, ~r/unknown :encode option :unknown/, fn ->
+        encode_body(%{a: 1}, encode: :unknown)
+      end
+    end
+
+    defp encode_body(body, opts) do
+      %Tesla.Env{body: body}
+      |> Tesla.Middleware.FormUrlencoded.encode(opts)
```
enilsen16, line 316: This directly calls `Tesla.Middleware.FormUrlencoded.encode/2`. Could you confirm `encode/2` is a public function (not just accessible via some test helper trick)? If it's already public API that's fine, but if it's private this test would fail — or silently exercise an internal function that could change without a semver bump.

lib/tesla/middleware/form_urlencoded.ex
tsitsiridakisiosif: nice to know, i removed it tho anyway!!

test/tesla/middleware/form_urlencoded_test.exs
```diff
@@ -124,6 +124,433 @@ defmodule Tesla.Middleware.FormUrlencodedTest do
...
+
+    test "unknown sub-option key raises" do
+      assert_raise ArgumentError, ~r/unknown option\(s\) \[:weird\]/, fn ->
+        %Tesla.Env{body: %{"k" => true}}
+        |> Tesla.Middleware.FormUrlencoded.encode(encode: {:brackets, weird: true})
+      end
+    end
+
+    defp sort_pairs(""), do: ""
+    defp sort_pairs(encoded), do: encoded |> String.split("&") |> Enum.sort() |> Enum.join("&")
```
enilsen16, line 503: Non-blocking nit: this sorts the `&`-split pairs lexicographically, so it assumes single-digit list indices. The current corpus tops out at ~5 elements so it's fine today, but if anyone later adds a fixture with 10+ list items, `foo[10]=...` would sort before `foo[2]=...` and the assertion would fail spuriously (a false failure, not a false pass — so it'd be caught, just confusingly). Could pre-empt it by splitting on the numeric index, but totally optional.

enilsen16 (approved):
Read through the encoder carefully and it's clean — nice work. The struct-clause-before-`is_map` ordering, the index-then-drop-`nil` trick for PHP-style sparse arrays, and threading `boolean_as` as a pure accumulator are all the right calls, and the PHP `http_build_query` parity corpus is a great way to pin the wire format. One tiny non-blocking note inline on a test helper.

PR #879, opened 2026-05-27
feat(finch-adapter): Finch adapter will pass through all supported options

aglassman:
The Finch adapter will now pass through all supported request and build options.

I initially did this because I'd like to pass through `pool_tag`, but figured I'd just update to support all the Finch request and build opts as of `finch:0.22.0`.

https://hexdocs.pm/finch/Finch.html#t:request_opt/0
https://hexdocs.pm/finch/Finch.Request.html#t:build_opt/0

### DIFF
--- a/lib/tesla/adapter/finch.ex
+++ b/lib/tesla/adapter/finch.ex
@@ -46,13 +46,27 @@ if Code.ensure_loaded?(Finch) do
         + `nil` or not specified - Responds without streaming using
           `Finch.request/3`.
 
-    ## [Finch options](https://hexdocs.pm/finch/Finch.html#request/3)
+    ## [Finch build options](https://hexdocs.pm/finch/Finch.html#build/5)
+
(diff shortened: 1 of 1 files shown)


yordis (approved):

yordis:
I wonder if we should do the opposite and exclude things that are related to Tesla itself 🤔

yordis:
@aglassman thank you so much, do you mind fixing CI?

aglassman:
@yordis Will do! I moved it to draft because there was testing I still needed to do.  

> I wonder if we should do the opposite and exclude things that are related to Tesla itself 🤔

Yeah, that could be a better option.  Although if there were ever named option overlap, that could be frustrating. The range over Finch versions is pretty wide 0.13.x-0.22.x currently supported.

yordis:
We can leave a TODO behind as well for now, they are close to a decade decisions that maybe a v2 would be granted at some point 😄

PR #880, opened 2026-05-31
feat(hackney)!: support hackney 4.x

tomekowal:
Adds support for Hackney 4.x.

There are two workarounds in this PR that are further explained in comments:
- I kept Hackney 1.x in mix.lock because Hackney 4.x depends on quic which requires Erlang 26 which breaks the CI matrix; AFAIU, the only way to upgrade Hackney to 4.x in mix.lock would be to drop Erlang 25 from the CI matrix
- Second commit introduces a function that does nothing except correcting wrong typespec in hackney for dialyzer to pass.

I made comments in places that I believe require attention during review.


lib/tesla/adapter/hackney.ex
```diff
@@ -3,7 +3,7 @@ if Code.ensure_loaded?(:hackney) do
     @moduledoc """
     Adapter for [hackney](https://github.com/benoitc/hackney).
 
-    Remember to add `{:hackney, "~> 1.13"}` to dependencies (and `:hackney` to applications in `mix.exs`)
+    Remember to add `{:hackney, "~> 4.0"}` to dependencies (and `:hackney` to applications in `mix.exs`)
```
tomekowal, line 6: I thought about leaving 1.13 or 4.0 in the docs, but in the end, it might better to always point to the latest version,

lib/tesla/adapter/hackney.ex
```diff
@@ -27,11 +27,16 @@ if Code.ensure_loaded?(:hackney) do
 
     ## Adapter specific options
 
-    - `:max_body` - Max response body size in bytes. Actual response may be bigger because hackney stops after the last chunk that surpasses `:max_body`.
+    - `:max_body` - Max response body size in bytes.
+      Works for hackney 1.x and 4.x when streaming the response.
```
tomekowal: I hope this disclaimer is clear. `max_body` worked in 1.x for both sych and async responses. In 4.x the option does not exist, but it was easy to preserve it by getting the body chunk by chunk and making sure we did not cross the line.

I also find `max_body` name a little unspecific. I'd rather have `max_response_body`, but it is also like that in other adapter, so that would be a breaking change.

We can also argue that it never pertains to request body because that on is under our control, so I am leaving it as is.

lib/tesla/adapter/hackney.ex
```diff
@@ -27,11 +27,16 @@ if Code.ensure_loaded?(:hackney) do
...
+      Actual response may be bigger
+      because hackney stops after the last chunk that surpasses `:max_body`.
     """
     @behaviour Tesla.Adapter
     alias Tesla.Multipart
 
+    defguardp is_hackney_ref(ref) when is_reference(ref) or is_pid(ref)
```
tomekowal: Hackney 1.x used reference and 4.x uses pid. Fortunately, we can reuse the code by checking if it is one or the other.

lib/tesla/adapter/hackney.ex
```diff
@@ -77,8 +83,10 @@ if Code.ensure_loaded?(:hackney) do
 
     defp request_stream(method, url, headers, body, opts) do
       with {:ok, ref} <- :hackney.request(method, url, headers, :stream, opts) do
-        case send_stream(ref, body) do
-          :ok -> handle(:hackney.start_response(ref), opts)
+        with :ok <- send_stream(ref, body),
+             :ok <- :hackney.finish_send_body(ref) do
```
tomekowal, line 103: I am not sure how it worked before without `finish_send_body`. The function existed in 1.x and it seems it should be called here always, but wasn't.

4.x is more strict about it.

lib/tesla/adapter/hackney.ex
```diff
@@ -99,18 +107,41 @@ if Code.ensure_loaded?(:hackney) do
...
+      with {:ok, body} <- read_response_body(ref, Keyword.get(opts, :max_body, :infinity)) do
         {:ok, status, headers, body}
       end
     end
 
     defp handle({:ok, status, headers, body}, _opts), do: {:ok, status, headers, body}
 
+    defp read_response_body(ref, :infinity), do: :hackney.body(ref)
+
+    defp read_response_body(ref, max) when is_integer(max) do
```
tomekowal, line 140: Here I am manually reproducing `max_body` for 4.x response streaming.

The alternative would be to drop it and crash or warn on using `max_body` with 4.x. Let me know if you'd like me to change it.

test/tesla/adapter/hackney_test.exs
```diff
@@ -33,23 +33,32 @@ defmodule Tesla.Adapter.HackneyTest do
...
-      url: "#{@http}/post",
-      body: body
-    }
-
-    assert {:ok, %Env{} = full_response} = call(request, with_body: true)
-    assert {:ok, %Env{} = limited_response} = call(request, with_body: true, max_body: 100)
-    assert full_response.status == 200
-    assert limited_response.status == 200
-    assert byte_size(limited_response.body) < byte_size(full_response.body)
+  @hackney_major :hackney
```
tomekowal: I didn't want to get rid of the test for `max_body` without streaming, so I put it behind version gate.

Let me know if you'd rather delete it.

mix.exs
```diff
@@ -56,7 +56,9 @@ defmodule Tesla.Mixfile do
 
       # http clients
       {:ibrowse, "4.4.2", optional: true},
-      {:hackney, "~> 1.21", optional: true},
+      # hackney in versions 2, 3 and up to 4.0.1 did not export all necessary functions
+      # including :hackney.body/1,2 and :hackney.stream_body/1,2
+      {:hackney, "~> 1.21 or ~> 4.0  and >= 4.0.2", optional: true},
```
tomekowal: I needed to check docs, but it seems `and` has higher precedence than or, so it is actually:

`~> 1.21 or (~> 4.0  and >= 4.0.2)`

mix.lock
```diff
@@ -18,7 +18,8 @@
...
+  "h2": {:hex, :h2, "0.6.1", "16a8cf76dc017aba1e47ea20bc928ae309892d36a59ca0cb43640bcabb2a270f", [:rebar3], [], "hexpm", "27fb7482341431dc9e1d96d2063c23664a52ec0ce70ba2fdaddb945356a51388"},
+  "hackney": {:hex, :hackney, "1.25.0", "390e9b83f31e5b325b9f43b76e1a785cbdb69b5b6cd4e079aa67835ded046867", [:rebar3], [{:certifi, "~> 2.15.0", [hex: :certifi, repo: "hexpm", optional: false]}, {:idna, "~> 6.1.0", [hex: :idna, repo: "hexpm", optional: false]}, {:metrics, "~> 1.0.0", [hex: :metrics, repo: "hexpm", optional: false]}, {:mimerl, "~> 1.4", [hex: :mimerl, repo: "hexpm", optional: false]}, {:parse_trans, "3.4.1", [hex: :parse_trans, repo: "hexpm", optional: false]}, {:ssl_verify_fun, "~> 1.1.0", [hex: :ssl_verify_fun, repo: "hexpm", optional: false]}, {:unicode_util_compat, "~> 0.7.1", [hex: :unicode_util_compat, repo: "hexpm", optional: false]}], "hexpm", "7209bfd75fd1f42467211ff8f59ea74d6f2a9e81cbcee95a56711ee79fd6b1d4"},
```
tomekowal: I have a dependency issue. Hackney 4.x requires quic that requires Erlang 26.

This breaks CI matrix which also runs on OTP 25. I am not sure how to proceed with that.

Do you want to drop Erlang 25 from the Matrix?

For now, I downgraded Hackney back to 1.x to make CI green.

tomekowal:
I've changed the PR title, but the action did not rerun. @yordis Could you rerun it?

test/tesla/adapter/hackney_test.exs
```diff
@@ -33,23 +33,32 @@ defmodule Tesla.Adapter.HackneyTest do
...
+                 |> Application.spec(:vsn)
+                 |> to_string()
+                 |> String.split(".")
+                 |> hd()
+                 |> String.to_integer()
+
+  if @hackney_major == 1 do
```
yordis: could we use `Version.compare/2` for these version checks?

lib/tesla/adapter/hackney.ex
yordis: Could you leave behind a TODO comments with all the details to remove eventually?

yordis:
@tomekowal thank you so much, I wonder, should we leave TODO comments with the proper context in the code to later remove 1.0 support?

tomekowal:
Do we want to remove 1.0 support? I don't believe keeping both in the same adapter adds that much complexity. There is not a single `if` statement necessary. The only small thing is checking if hackney reference is either Erlang's reference or PID.

It would be totally fine to leave both 1.x and 4.x support indefinitely.

halostatue:
As someone adjacent to security…I don't think keeping 1.0 is a good idea. It's not getting any more updates and has three unresolved CVEs.

But that changes a number of things for consumers including the minimum supported OTP.

yordis:
I am in a tricky situation here, it is effectively a breaking change for Tesla ... 😭 What do I do? Maybe I shouldn't feel so emotionally attach to cut a v2 or put a big disclaimer in the release note?

tomekowal:
I think that makes sense for a major upgrade:
- cut the support for OTP 25
- cut support for Hackney 1.x
- any other stuff that you would like to do but compatibility held you back?

yordis:
use `feat(hackney)!` the `!` is the key here, lets break changes 😢

tomekowal:
I've change the title. Would you like to me to make other changes that drop Hackney 1.x support completely?

yordis:
yeah, drop it completely by now

tomekowal:
OK, so:
- I've removed traces of Hackney v1
- Removed Erlang 25 from test matrix
- updated excoveralls in gun lockfile (because updating hackney caused errors)
- the Semantic PR Title failure seems to be not from the title itself but from how the job works

yordis (approved):

yordis:
@tomekowal thank you so much for the help!

tomekowal:
Happy to help :)

PR #882, opened 2026-06-02
fix(hackney): restore hackney 1.x support

yordis:
- PR #880 was merged as a breaking change, but hackney 1.x and 4.x can coexist behind a single guard; users on 1.x should not see a spurious major bump.
- `Release-As: 1.18.3` in the commit footer overrides release-please so the next release is a patch, not a major.

### DIFF
--- a/lib/tesla/adapter/hackney.ex
+++ b/lib/tesla/adapter/hackney.ex
@@ -3,7 +3,7 @@ if Code.ensure_loaded?(:hackney) do
     @moduledoc """
     Adapter for [hackney](https://github.com/benoitc/hackney).
 
-    Remember to add `{:hackney, "~> 4.0"}` to dependencies (and `:hackney` to applications in `mix.exs`)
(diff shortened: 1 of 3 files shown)


yordis:
@tomekowal @halostatue hey folks, I had to revert to support < 4.0 (please @tomekowal check again just in case I made a mistake here) since I need to syncronize this with some work from CVEs cc: @maennchen

I agree with @halostatue on avoiding supporting broken ones, but, they are some consequences here that I can't simply do it without affecting far too many moving pieces

maennchen:
We need to keep support while we fix some other CVEs since there's this issue: https://github.com/joedevivo/chatterbox/issues/154

I don't want the hackney update to prevent users from updating the CVEs that are being merged shortly.

Afterwards we can also directly drop support and release a new major version only supporting v4.

tomekowal:
👍 That looks right

PR #886, opened 2026-06-04
feat: support {:form,data} tagged body tuples

tsitsiridakisiosif:


lib/tesla/middleware/form_urlencoded.ex
```diff
@@ -219,11 +219,13 @@ defmodule Tesla.Middleware.FormUrlencoded do
     end
   end
 
+  defp encodable?(%{body: {:form, _}}), do: true
```
yordis: should the name be `:form_urlencoded`?

yordis (approved):

PR #891, opened 2026-06-25
chore: drop archived exjsx JSON engine

gilbertwong96:
The `exjsx` JSON library is archived and no longer maintained, so drop direct support for it as an engine in Tesla:

- remove the optional `exjsx` dependency
- remove the `engine: JSX` example and update the `:engine` doc in `Tesla.Middleware.JSON`
- replace the remaining `decode: &JSX.decode/1, encode: &JSX.encode/1` doc example with `&Jason.decode/1` / `&Jason.encode/1` (a maintained engine) to keep the `:decode` / `:encode` function example
- remove the `Engine: exjsx` test block

`exjsx` remains in the lockfile only as a transitive dependency of the test-only `httparrot` server. The Elixir 1.20.1 compilation warning fixes (preferred_cli_env, unused require, gun adapter) are split into a separate PR.

### DIFF
--- a/lib/tesla/middleware/json.ex
+++ b/lib/tesla/middleware/json.ex
@@ -35,11 +35,9 @@ defmodule Tesla.Middleware.JSON do
         # or
         {Tesla.Middleware.JSON, engine: JSON}
         # or
-        {Tesla.Middleware.JSON, engine: JSX, engine_opts: [strict: [:comments]]},
-        # or
         {Tesla.Middleware.JSON, engine: Poison, engine_opts: [keys: :atoms]},
         # or
-        {Tesla.Middleware.JSON, decode: &JSX.decode/1, encode: &JSX.encode/1}
+        {Tesla.Middleware.JSON, decode: &Jason.decode/1, encode: &Jason.encode/1}
       ])
     end
   end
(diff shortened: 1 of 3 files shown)


yordis:
Hey @gilbertwong96 thank you so much for the contribution. They are few things happening in this PR, ideally, we can keep the changes around a singular subject.

About `hackney` I understand your intent here, but I can't force everyone to upgrade, see https://github.com/elixir-tesla/tesla/pull/882#issuecomment-4605518423

lib/tesla/middleware/json.ex
```diff
@@ -35,11 +35,7 @@ defmodule Tesla.Middleware.JSON do
         # or
         {Tesla.Middleware.JSON, engine: JSON}
         # or
-        {Tesla.Middleware.JSON, engine: JSX, engine_opts: [strict: [:comments]]},
-        # or
-        {Tesla.Middleware.JSON, engine: Poison, engine_opts: [keys: :atoms]},
-        # or
-        {Tesla.Middleware.JSON, decode: &JSX.decode/1, encode: &JSX.encode/1}
+        {Tesla.Middleware.JSON, engine: Poison, engine_opts: [keys: :atoms]}
```
yordis: Not sure the examples were removed, can you expand upon your reasoning here?

gilbertwong96: exjsx is already an archived project, I don't think it's good idea to still support archived exjsx for this project. And if someone don't want to upgrade, they should keep using the legacy version of tesla rather than upgrade tesla to the latest version. If they want to upgrade Tesla to the latest version, they should upgrade along with other dependencies as well.

gilbertwong96:
> Hey @gilbertwong96 thank you so much for the contribution. They are few things happening in this PR, ideally, we can keep the changes around a singular subject.
> 
> About `hackney` I understand your intent here, but I can't force everyone to upgrade, see [#882 (comment)](https://github.com/elixir-tesla/tesla/pull/882#issuecomment-4605518423)

Sure, I will split them into different PRs later.

gilbertwong96:
Just force pushed

gilbertwong96:
Opened #892 for silence warnings in Elixir 1.20

yordis (approved):

PR #913, opened 2026-08-16
fix(finch-adapter): surface the reason a streamed response stopped

### LINKED ISSUE #912: Finch adapter discards the reason a streamed response stopped
We ran into this while consuming a Server-Sent Events endpoint. It took us a while to work out what was happening, so I thought it was worth writing up.

When the Finch adapter streams a response, the body stream stops the same way whether the response finished normally, the connection broke partway through, or we timed out waiting for the next chunk. All three return `nil` from the `Stream.unfold`, and in the error case the reason is bound to an unused variable and thrown away.

```elixir
Stream.unfold(nil, fn _ ->
  receive do
    {^ref, {:data, data}} ->
      {data, nil}

    {^ref, :eof} ->
      Task.await(task)
      nil

    {^ref, {:error, _error}} ->
      Task.shutdown(task, :brutal_kill)
      nil
  after
    opts[:receive_timeout] ->
      Task.shutdown(task, :brutal_kill)
      nil
  end
end)
```

The outer `receive` does tell these apart, so anything that fails before the headers arrive comes back as an error with a reason. It is only once headers have arrived that the reason goes missing.

For us that meant a half finished response and a healthy short one looked identical. Our replies end with a terminal event, so we could see that the last event never arrived, but not whether the server had closed early, the connection had dropped, or we had given up waiting. All of those reached people as the same message, and working out which one had actually happened meant reading the other service's source and guessing. It also makes retries hard to reason about, since a timeout is usually worth retrying and a genuinely short response usually is not.

While reading the adapter I noticed something that may be related. `req_opts` is built by taking only `:pool_timeout` and `:receive_timeout`, so `:request_timeout` never reaches Finch. On HTTP/1 that is the only setting bounding the total length of a request, since `:receive_timeout` applies per read, so as far as I can tell there is currently no way to cap a long streamed response through this adapter. Happy to move that into its own issue if you would prefer.

I am not sure what the right shape of a fix is. Putting a sentinel value into the stream is the obvious idea, but it would break anything that assumes the elements are binaries, including your own SSE middleware, which concatenates them. Handing the caller a callback, or recording the reason somewhere it can be read after the stream is consumed, both work but neither feels clearly right to me. Returning tagged tuples behind an option would be tidier but is a bigger change.

We are running a patched copy of the adapter for now, which we would rather not keep doing. If you have a preference for the approach, I am glad to put a PR together.

Tesla 1.18.3, Finch 0.23.0, Mint 1.9.3, Elixir 1.18.3 on OTP 27. Still present on `master` as far as I can see.

yordis:
Closes #912

- A streamed response that broke halfway through was indistinguishable from one that finished, so callers accepted truncated data as a complete answer.
- A dropped connection and an exhausted `:receive_timeout` need to be told apart to decide whether retrying is worthwhile.
- Failures after the headers arrived cannot be returned as `{:error, reason}`, since the request already succeeded from the caller point of view, so they are raised while the body is consumed instead.
- Keeping stream elements as binaries leaves `Tesla.Middleware.SSE` and any other consumer that concatenates chunks untouched.
- Matches how the Mint adapter already reports mid-stream failures.

### DIFF
--- a/lib/tesla/adapter/finch.ex
+++ b/lib/tesla/adapter/finch.ex
@@ -46,6 +46,28 @@ if Code.ensure_loaded?(Finch) do
         + `nil` or not specified - Responds without streaming using
           `Finch.request/3`.
 
+    ## Streamed response failures
+
+    When `response: :stream` is used, failures that happen after the response
+    headers arrived cannot be returned as `{:error, reason}`, because the
+    request already succeeded from the caller point of view. Such failures raise
+    `Tesla.Error` while the body stream is being consumed, so that a truncated
+    response is never mistaken for a complete one:
+
+    ```elixir
+    try do
+      Enum.each(env.body, &handle_chunk/1)
+    rescue
+      e in Tesla.Error ->
+        # e.reason is, for example, :closed or :timeout
+        {:error, e.reason}
+    end
+    ```
+
+    Waiting longer than `:receive_timeout` for the next chunk raises with reason
+    `:timeout`. A server that ends the response early without failing the
+    connection is not an error, and the stream simply ends.
+
     ## [Finch build options](https://hexdocs.pm/finch/Finch.html#build/5)
 
       * `:unix_socket` - Path to a Unix domain socket to connect to instead of the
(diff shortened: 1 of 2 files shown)


yordis:
@elias-ba can you give it a try to this one?

elias-ba:
Tried it at 9045218 against the case from #912, an SSE stream consumed through `Tesla.Middleware.SSE`:

- stalls past `:receive_timeout` gives `Tesla.Error` with reason `:timeout`
- server hangs up mid stream gives `Tesla.Error` with reason `:closed`
- clean end just finishes, events intact

That is what I needed. Telling those two apart through the SSE middleware was the part I could not do before, since it concatenates chunks as binaries and a sentinel inside the stream was never going to work.

Thanks for turning it around so quickly @yordis this is amazing

